PR #54, opened 2015-02-27
Test against latest Ruby 2.1 & 2.2

nTraum:
Hi there,

this PR includes 3 things regarding the build environment:

* Run specs against latest Ruby 2.1 (2.1.5 instead of 2.1.1)
* Add support for Ruby 2.2
* Drop support for Ruby 1.9.3

We could also think about removing support for MRI 1.9.3 as official support has ended just recently: https://www.ruby-lang.org/en/news/2015/02/23/support-for-ruby-1-9-3-has-ended/.

@jstorimer, @wvanbergen looking forward to your feedback. :)

Cheers

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,11 +1,11 @@
 language: ruby
 rvm:
-  - 1.9.3
-  - 2.0.0
-  - 2.1.1
-  - ruby-head
+  - 2.0
+  - 2.1
+  - 2.2
   - jruby-19mode
   - rbx-2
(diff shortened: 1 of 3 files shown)


wvanbergen:
This looks good! We can drop support for ruby 1.9 as well.

nTraum:
Removed 1.9.3 from Travis as suggested.

wvanbergen:
Thanks!

PR #58, opened 2015-05-06
OpenTSDB backend support with tags

mikwat:
The StatsD [OpenTSDB backend](https://github.com/danslimmon/statsd-opentsdb-backend) supports tags in a different format than the existing Datadog implementation.

Feedback welcome @jstorimer @wvanbergen


mikwat:
@jstorimer @wvanbergen Feedback welcome whenever you have a chance to look at this. Thanks.

wvanbergen:
I'll try to look at this this week.

Note: at this point, it may make sense to start splitting up different implementations in different classes, instead of a bunch of if statements. This code is in a hot path (at least for us), and I rather run the if statement during initialization instead of every StatsD packet.

mikwat:
That makes sense, @wvanbergen I might be able to take a pass at splitting it up.

mikwat:
Hey @wvanbergen, I've split out the OpenTSDB logic into a separate class that inherits from `UDPBackend`. If you like this approach, maybe something similar could be done for the other implementations. What do you think?

PR #60, opened 2015-06-29
metric: replacing :: is not needed for statsd protocol

Sirupsen:
Rice up `normalize_tags` even harder. With e.g. the Datadog statsd implementation it's not actually required to replace colons, they can be part of tags since keys and values are seperated by the first colon—different tags by colon. I think this is reasonable to do.

What do you think @wvanbergen ?

### DIFF
--- a/test/metric_test.rb
+++ b/test/metric_test.rb
@@ -26,7 +26,7 @@ class MetricTest < Minitest::Test
 
   def test_handle_bad_tags
     assert_equal ['ignored'], StatsD::Instrument::Metric.normalize_tags(['igno|red'])
-    assert_equal ['lolclass:omglol'], StatsD::Instrument::Metric.normalize_tags({ :"lol::class" => "omg::lol" })
+    assert_equal ['lol::class:omg::lol'], StatsD::Instrument::Metric.normalize_tags({ :"lol::class" => "omg::lol" })
   end
 
   def test_rewrite_tags_provided_as_hash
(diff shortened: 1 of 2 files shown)


test/metric_test.rb
```diff
@@ -26,7 +26,7 @@ def test_name_prefix
 
   def test_handle_bad_tags
     assert_equal ['ignored'], StatsD::Instrument::Metric.normalize_tags(['igno|red'])
-    assert_equal ['lolclass:omglol'], StatsD::Instrument::Metric.normalize_tags({ :"lol::class" => "omg::lol" })
+    assert_equal ['lol::class:omg::lol'], StatsD::Instrument::Metric.normalize_tags({ :"lol::class" => "omg::lol" })
```
wvanbergen: SO how will this show up in DD? `lol` => `:class:omg::lol`? Or `lol::class` => `omg::lol`?

Sirupsen: The former, there will be a tag named `lol` with the value `:class:omg::lol`.

Sirupsen: From what I've seen in the wild this is OK. The tag name is almost always hardcoded, and the value is sometimes e.g. a class converted to a string.

wvanbergen: Yeah, as long as the value will be fine (which can be a Ruby class name with namespacing), I am OK with this :+1:

wvanbergen:
:shipit:

Sirupsen:
@wvanbergen :clap: mind merging and releasing? :-)

PR #61, opened 2015-07-24
Ensure metric name(s) always appears in assert_no_statsd_calls assertion message

christhomson:
r: @wvanbergen 

I tried using `assert_no_statsd_calls` with no `metric_name` argument in a test today, and a statsd call *was* made,  but when the assertion wasn't met I was presented with a rather unhelpful error message: `No StatsD calls for metric  expected.` I had no idea which metric was being called unexpectedly.

This PR changes the assertion message to show metric names that were actually called unexpectedly.


wvanbergen:
nice :+1:

wvanbergen:
Released as 2.0.9

christhomson:
Thanks! :heart:

PR #62, opened 2015-09-16
assertion options[:ignore_tags]

fw42:
This adds `options[:ignore_tags]` to the StatsD test assertions, allowing you to add a list (or hash) of tags that are allowed to be present even though they weren't expected (but if they were expected then they have to be present).

The splitting thing seems a bit hacky, but not sure how else to do it.

@wvanbergen, thoughts?

### DIFF
--- a/lib/statsd/instrument/assertions.rb
+++ b/lib/statsd/instrument/assertions.rb
@@ -43,7 +43,22 @@ module StatsD::Instrument::Assertions
 
     assert_equal options[:sample_rate], metric.sample_rate, "Unexpected value submitted for StatsD metric #{metric_name}" if options[:sample_rate]
     assert_equal options[:value], metric.value, "Unexpected StatsD sample rate for metric #{metric_name}" if options[:value]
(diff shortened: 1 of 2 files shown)


lib/statsd/instrument/assertions.rb
```diff
@@ -43,7 +43,22 @@ def assert_statsd_call(metric_type, metric_name, options = {}, &block)
...
+      if options[:ignore_tags]
+        ignored_tags = Set.new(StatsD::Instrument::Metric.normalize_tags(options[:ignore_tags])) - expected_tags
+        actual_tags -= ignored_tags
+
+        if options[:ignore_tags].is_a?(Array)
+          actual_tags.delete_if{ |key| options[:ignore_tags].include?(key.split(":").first) }
+        end
+      end
+
+      assert_equal expected_tags, actual_tags, "Unexpected StatsD tags for metric #{metric_name}"
```
wvanbergen: While we are at it, can we also make this error message more useful?

wvanbergen:
I think this is a good addition. I don't really have a suggestion how to do the splitting in a better way, so I am fine with this implementation :+1:

PR #63, opened 2015-10-02
Fix mixup of assertion failure messages for value and sample_rate

Soleone:
Looks like the failure messages were accidentally mixed up.

for review @wvanbergen


wvanbergen:
:+1:

PR #65, opened 2015-10-05
Don't change method visibility scope

gmalette:
`StatsD::Instrument` defines all methods public. It should respect the initial visibility of the method. 

ping @wvanbergen @jstorimer

### DIFF
--- a/lib/statsd/instrument.rb
+++ b/lib/statsd/instrument.rb
@@ -210,9 +210,20 @@ module StatsD
       raise ArgumentError, "already instrumented #{method} for #{self.name}" if method_defined? method_name_without_statsd
       raise ArgumentError, "could not find method #{method} for #{self.name}" unless method_defined?(method) || private_method_defined?(method)
 
+      method_scope = case
+      when private_method_defined?(method)
+        :private
+      when protected_method_defined?(method)
+        :protected
+      else
+        :public
+      end
+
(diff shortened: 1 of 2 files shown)


gmalette:
cc @sgn

wvanbergen:
You mean when using the metaprogramming to instrument an existing method?

gmalette:
> You mean when using the metaprogramming to instrument an existing method?

Yes. I believe the tests demonstrate the issue pretty well. 

```ruby
class A
  extend StatsD::Instrument
  private

  def toto
  end
end

A.new.toto
#=> NoMethodError: private method `toto' called for #<A:0x007fd8327dfe30>

A.statsd_count :toto, "test"

# A#toto is now public because it's a new, public method
A.new.toto
=> nil
```

wvanbergen:
OK, makes sense. I will see if I can fix this :+1:

gmalette:
@wvanbergen what's wrong with this PR?

wvanbergen:
AH, this is a PR. I thought this was an issue :P

wvanbergen:
:+1:

gmalette:
Worth bumping the major? I think this will break some method calls

wvanbergen:
I don't know if this warrants a major version break; the previous behaviour was never intended nor advertised. I don't expect anybody was using this gem to change their methods from private to public.

PR #66, opened 2015-10-19
Don't create new method with public visibility scope

tjoyal:
Proposed extra visibility tune-up on top of #65 

I do not think there is any reasons for dynamically created methods `method_name_without_statsd` and `method_name_with_statsd` to be generated in the public scope.

At first I tried to bing the scope to the caller scope, but turns out some projects are already using this gem in their initialization this way, leading to a public scope which I think is not right.
```ruby
Object.extend StatsD::Instrument
Object.statsd_measure :call, 'Object.call'
```

Also, found and fix an issue where removing a StatsD would not always restore the appropriate scope. (Should I make a PR just for that?)

Please review @gmalette @wvanbergen

### DIFF
--- a/lib/statsd/instrument.rb
+++ b/lib/statsd/instrument.rb
@@ -210,7 +210,32 @@ module StatsD
       raise ArgumentError, "already instrumented #{method} for #{self.name}" if method_defined? method_name_without_statsd
       raise ArgumentError, "could not find method #{method} for #{self.name}" unless method_defined?(method) || private_method_defined?(method)
 
-      method_scope = case
+      method_scope = method_visibility(method)
+
+      alias_method method_name_without_statsd, method
+      yield method_name_without_statsd, method_name_with_statsd, metric_name
+      alias_method method, method_name_with_statsd
+
+      private(method_name_without_statsd)
+      send(method_scope, method)
+      private(method_name_with_statsd)
+    end
+
+    def remove_from_method(method, name, action)
+      method_name_without_statsd = :"#{method}_for_#{action}_on_#{self.name}_without_#{name}"
+      method_name_with_statsd = :"#{method}_for_#{action}_on_#{self.name}_with_#{name}"
+
+      method_scope = method_visibility(method)
+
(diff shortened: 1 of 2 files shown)


lib/statsd/instrument.rb
```diff
@@ -210,28 +210,39 @@ def add_to_method(method, name, action, &block)
...
 
-      method_scope = case
-      when private_method_defined?(method)
-        :private
-      when protected_method_defined?(method)
-        :protected
-      else
-        :public
-      end
+      method_scope = visibility_scope_for(method)
```
gmalette: `s/visibility_scope_for/method_visibility`

gmalette:
:+1:

wvanbergen:
What is the visibility of the alias methods? (e.g. `method_without_statsd`)

tjoyal:
> What is the visibility of the alias methods? (e.g. `method_without_statsd`)

My tests (and the code I'm trying to fix with this PR) showed that even if the helper method (eg.: `statsd_measure`) is called from a `private` scope, since the helper itself is `public` the `define_method` in it is run within that scope.

I'm fairly confident but do not have past experience with alias methods and scopes. Would you prefer to add an additional person to the discussion?

tjoyal:
ping :)

wvanbergen:
I am OK with this but I'd like to ping somebody with Rubt expertise on this one.

wvanbergen:
Fixes https://github.com/Shopify/cloud/issues/175

tjoyal:
@sgnr can I have your Rubt expertise on this please :smile_cat:

sgnr:
If I understand correctly, you question is: which visibility is given to method created through `define_method`?

sgnr:
If so, it is always `public`: https://github.com/ruby/ruby/blob/d9ac64fdc79ec2aefe5ed21b43b151caa296c5e3/class.c#L1487..L1491

tjoyal:
But are you ok with the PR itself? :)

I do not really want them to be public

sgnr:
Yes that looks sound :raised_hands:

PR #70, opened 2015-10-27
Add InfluxDB 0.9+ with tag supports

v9n:
Hi,

Thanks for a great package.
### Short summary:

I'm adding a backend to send metric to InfluxDB. Since it's a new backend, and it simply extends `UDPBackend` to overwrite a method, I don't think it will affect performance that much for you guys, as you guys mention in other issue that this code is in a hot path.
### Detail

I want to add tag for supporting InfluxDB backend 0.9. The InfluxDB community usually uses this package https://github.com/bernd/statsd-influxdb-backend to forward metric from StatsD->InfluxDB. So the influxdb backend in fully compatible with statsd. Because https://github.com/bernd/statsd-influxdb-backend works as a statsd backend so it has capable of flusing metric into InfluxDB and client doesn't need to know about InfluxDB. 

The issues is, StatsD doesn't support tag, while as InfluxDB does. Some efforts was made here. But all of them will require StatsD to change the structure of metric for appending tag like `metricname:value|type|@sample_rate` to `metricname:value|type|@sample_rate|#tag1=val1,tag2=val2` similar to data dog. That's won't happen soon. 

So I take another approach by moving the tag like this: 

`metricname#tag1=val1,tag2=val2:value|type|@sample_rate`

I will then have the statsd-influxdb-backend parse metric name to write tags into InfluxDB.


kureikain:
Hi @wvanbergen any though on this backend?

arthurnn:
Not sure this should be part of the main gem. 
You probably can easily have this extension in your app, and even do a third-party gem for it. But I would say the gem should have what 80% of the users need.

lib/statsd/instrument/backends/influxdb_backend.rb
```diff
@@ -0,0 +1,23 @@
...
+  # metricname#tag1=val1,tag2=val2:metricvalue|metrictype|@sample
+  # It's very similar to datadog backend, however, it move the tags into metric
+  # name. The backend where stat is flushed to is responsible to parse metric
+  # name for getting tag values
+  class InfluxDBBackend < StatsD::Instrument::Backends::UDPBackend
+
+    def generate_packet(metric)
+      command = "#{metric.name}"
+      if metric.tags
+        command << "##{metric.tags.join(',')}"
```
fbbergamo: Maybe this class should be responsible for escaping values if its a string.
https://docs.influxdata.com/influxdb/v0.9/write_protocols/write_syntax/#escaping-characters

wvanbergen:
Thank you for your contribution. We are currently in the process of doing a major release that completely changes the way we support different implementations. The CHANGELOG contains more information about the next major version.

We will not be accepting PRs that add support for other implementations on the legacy implementation as a result. Instead, to support InfluxDB in the new `StatsD::Instrument::Client` implementation, you'll have to implement a subclass of the `DatagramBuilder` class.

wvanbergen:
Thanks for your contribution. I will close this, as they will require a full rewrite to be compatible with v3 of the library like I mentioned in my previous comment. I am happy to support anyone that wants to add InfluxDB support to the new version if anyone is willing to work on it.

PR #71, opened 2015-10-30
Support for dynamic metric options

adamgotterer:
Specifically implemented and tested for using a lambda as a tag value which will take the caller and wrapped function args as params. The code is designed to support future metric options.

@jstorimer @wvanbergen

### DIFF
--- a/README.md
+++ b/README.md
@@ -196,6 +196,9 @@ The Datadog implementation support tags, which you can use to slice and dice met
 ``` ruby
 StatsD.increment('my.counter', tags: ['env:production', 'unicorn'])
 GoogleBase.statsd_count :insert, 'GoogleBase.insert', tags: ['env:production']
+
+# Use a lambda for generating dynamic tags from the caller or method arguments
+GoogleBase.statsd_count :insert, 'GoogleBase.insert', tags: lambda { |calle, args| [args[0]] }
 ```
 
 If implementation is not set to `:datadog`, tags will not be included in the UDP packets, and a
--- a/lib/statsd/instrument.rb
+++ b/lib/statsd/instrument.rb
@@ -168,10 +168,11 @@ module StatsD
     # @param name (see #statsd_measure)
     # @param metric_options (see #statsd_measure)
     # @return [void]
-    def statsd_count(method, name, *metric_options)
+    def statsd_count(method, name, *metric_option_args)
+      #metric_options = metric_options_args
       add_to_method(method, name, :count) do |old_method, new_method, metric_name|
         define_method(new_method) do |*args, &block|
-          metric_options = StatsD::Instrument.process_metric_options(metric_options, self, *args)
+          metric_options = StatsD::Instrument.process_metric_options(metric_option_args, self, *args)
           StatsD.increment(StatsD::Instrument.generate_metric_name(metric_name, self, *args), 1, *metric_options)
           send(old_method, *args, &block)
         end
--- a/lib/statsd/instrument.rb
+++ b/lib/statsd/instrument.rb
@@ -46,6 +46,21 @@ module StatsD
       metric_name.respond_to?(:call) ? metric_name.call(callee, args).gsub('::', '.') : metric_name.gsub('::', '.')
     end
 
+    # @private
+    def self.process_metric_options(metric_options, calle, *method_args)
+      metric_options.map do |m_value|
+        next m_value unless m_value.is_a?(Hash)
+
+        m_value.each_with_object({}) do |(name, value), obj|
+          if value.respond_to?(:call)
+            obj[name] = value.call(calle, method_args)
+          else
+            obj[name] = value
+          end
+        end
+      end
+    end
+
     if Process.respond_to?(:clock_gettime)
       # @private
       def self.duration
@@ -105,6 +120,7 @@ module StatsD
             result
           ensure
             suffix = truthiness == false ? 'failure' : 'success'
+            metric_options = StatsD::Instrument.process_metric_options(metric_options, self, *args)
             StatsD.increment("#{StatsD::Instrument.generate_metric_name(metric_name, self, *args)}.#{suffix}", 1, *metric_options)
           end
         end
@@ -136,6 +152,7 @@ module StatsD
             truthiness = (yield(result) rescue false) if block_given?
             result
           ensure
+            metric_options = StatsD::Instrument.process_metric_options(metric_options, self, *args)
             StatsD.increment(StatsD::Instrument.generate_metric_name(metric_name, self, *args), *metric_options) if truthiness
           end
         end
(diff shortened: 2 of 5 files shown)


adamgotterer:
Any feedback?

README.md
```diff
@@ -196,6 +196,9 @@ The Datadog implementation support tags, which you can use to slice and dice met
 ``` ruby
 StatsD.increment('my.counter', tags: ['env:production', 'unicorn'])
 GoogleBase.statsd_count :insert, 'GoogleBase.insert', tags: ['env:production']
+
+# Use a lambda for generating dynamic tags from the caller or method arguments
+StatsD.increment('my.counter', tags: lambda { |calle, args| ['dynamic tag'])
```
wvanbergen: I don't see where this is implemented.

I don't actually mind it not being there, because it's not useful outside of the metaprogramming methods and only makes the implementation more complex and slower. I'd like to keep the basic `StatsD.increment` call as fast as possible, because it is in many hot code paths.

wvanbergen:
I like this idea, but I am a bit worried about the performance implications. Any chance we can do a basic benchmark, e.g. by calling an instrumented method `n` times using the current master version vs. this implementation?

adamgotterer:
Fixed a small bug and pushed a new commit for it. 

Below is the benchmark code and results for 1 million iterations. They are arguably pretty nearly equal since the margin is so small. But we can conclusively say the dynamic tagging doesn't have noticeable or negative impact.

```ruby
# Benchmark code. Not the most pretty thing, but it works

require 'benchmark'

###########
# Comment this section out to use the system version of the gem
require 'bundler/setup'
Bundler.require('/Users/adamgotterer/sites/statsd-instrument')
##########

require 'statsd-instrument'

puts 'Running on version: ' + StatsD::Instrument::VERSION.to_s

ITERATIONS = 1_000_000

StatsD.logger = Logger.new('/dev/null')
StatsD.backend = StatsD::Instrument::Backends::CaptureBackend.new

class SomeClass
  def increment(rand_val)
  end
end

SomeClass.extend StatsD::Instrument

# SomeClass.statsd_count :increment, 'inc', tags: ['test']
SomeClass.statsd_count :increment, 'inc', tags: -> (_, args) { [args[0]] }

Benchmark.bm do |x|
  someclass = SomeClass.new

  x.report do
    ITERATIONS.times do
      r = rand(1..100)
      someclass.increment(r)
    end
  end
end

puts "Captured all increments: " + (StatsD.backend.collected_metrics.count == ITERATIONS).to_s
```

```
# Results:

Static Tags (2.0.10 master):

"Running on version: 2.0.10"
       user     system      total        real
   8.700000   0.180000   8.880000 (  8.907977)
"Captured all increments: true"

-----------------

"Running on version: 2.0.10"
       user     system      total        real
   8.720000   0.170000   8.890000 (  8.889871)
"Captured all increments: true"

-----------------

"Running on version: 2.0.10"
       user     system      total        real
   9.220000   0.170000   9.390000 (  9.404807)
"Captured all increments: true"

-----------------

"Running on version: 2.0.10"
       user     system      total        real
   8.930000   0.170000   9.100000 (  9.103763)
"Captured all increments: true"




Dynamic Tags (on dynamic tagging branch):

"Running on version: 2.0.10-tags"
       user     system      total        real
   9.360000   0.220000   9.580000 (  9.750842)
"Captured all increments: true"

-----------------

"Running on version: 2.0.10-tags"
       user     system      total        real
   8.940000   0.170000   9.110000 (  9.118334)
"Captured all increments: true"

-----------------

"Running on version: 2.0.10-tags"
       user     system      total        real
   8.850000   0.170000   9.020000 (  9.020030)
"Captured all increments: true"

-----------------

"Running on version: 2.0.10-tags"
       user     system      total        real
   8.890000   0.160000   9.050000 (  9.052533)
"Captured all increments: true"




Static Tags (on dynamic tagging):

"Running on version: 2.0.10-tags"
       user     system      total        real
   9.790000   0.230000  10.020000 ( 10.270697)
"Captured all increments: true"

-----------------

"Running on version: 2.0.10-tags"
       user     system      total        real
   8.990000   0.170000   9.160000 (  9.165445)
"Captured all increments: true"

-----------------

"Running on version: 2.0.10-tags"
       user     system      total        real
   8.870000   0.160000   9.030000 (  9.041330)
"Captured all increments: true"

-----------------

"Running on version: 2.0.10-tags"
       user     system      total        real
   9.000000   0.170000   9.170000 (  9.180854)
"Captured all increments: true"
``

File: README.md (hunk 1 above)
Comment (adamgotterer) [inline README.md]: Yup, sorry that was a silly mistake. Fixed.

**Comment by adamgotterer** [comment]
Rebased and force pushed readme fixes.

**Comment by adamgotterer** [comment]
Any thoughts or feedback?

File: lib/statsd/instrument/metric.rb [hunk 2]
```diff
@@ -103,6 +103,6 @@ def inspect
   def self.normalize_tags(tags)
     return unless tags
     tags = tags.map { |k, v| k.to_s + ":".freeze + v.to_s } if tags.is_a?(Hash)
-    tags.map { |tag| tag.tr('|,'.freeze, ''.freeze) }
+    tags.map { |tag| tag.to_s.tr('|,'.freeze, ''.freeze) }
```
wvanbergen: Why is this necessary? This is a very hot code path for us, and we'd like to not add any unnecessary processing here.

lib/statsd/instrument.rb
```diff
@@ -151,9 +168,11 @@ def statsd_count_if(method, name, *metric_options)
     # @param name (see #statsd_measure)
     # @param metric_options (see #statsd_measure)
     # @return [void]
-    def statsd_count(method, name, *metric_options)
+    def statsd_count(method, name, *metric_option_args)
+      #metric_options = metric_options_args
       add_to_method(method, name, :count) do |old_method, new_method, metric_name|
         define_method(new_method) do |*args, &block|
+          metric_options = StatsD::Instrument.process_metric_options(metric_option_args, self, *args)
           StatsD.increment(StatsD::Instrument.generate_metric_name(metric_name, self, *args), 1, *metric_options)
```
wvanbergen: Can you move these two lines into an `ensure` block, so it runs after the original method has run? This brings it in line with the other methods.

lib/statsd/instrument.rb
```diff
@@ -46,6 +46,21 @@ def self.generate_metric_name(metric_name, callee, *args)
...
     end
 
+    # @private
+    def self.process_metric_options(metric_options, calle, *method_args)
+      metric_options.map do |m_value|
+        next m_value unless m_value.is_a?(Hash)
+
+        m_value.each_with_object({}) do |(name, value), obj|
+          if value.respond_to?(:call)
+            obj[name] = value.call(calle, method_args)
```
wvanbergen: This library gives a strong guarantee that it never raises exceptions. We should really be catching exception here.

wvanbergen:
Closing this because it's stale. Feel free to open a new PR if this is still a desired change.

PR #75, opened 2015-11-26
[Matchers] Add compound expectation

haeky:
We needed to have composition with statsd-instruments matchers. We can now use

```ruby
expect(:add_potato).to trigger_statsd_increment('potato_count').and be > 10
```

ping @wvanbergen @mlhamel


mlhamel:
:+1: awesome, at one point we might want to release that stuff as a new release and updating the changelog but it is not really needed in that PR

wvanbergen:
LGTM.

PR #80, opened 2016-04-08
tags+times bug

fw42:
I think I found a small and subtle bug in the StatsD assertion helper.

When submitting the same metric twice with different tags, the assertion helper with `times: 2` will not complain, even if only one of the metrics has the expected tag.

@wvanbergen, do you agree that this is a bug? If so, @surrahman, can you look into a fix?

### DIFF
--- a/test/assertions_test.rb
+++ b/test/assertions_test.rb
@@ -81,6 +81,20 @@ class AssertionsTest < Minitest::Test
     end
 
     assert_no_assertion_triggered do
+      @test_case.assert_statsd_increment('counter', times: 2, tags: ['foo:1']) do
+        StatsD.increment('counter', tags: { foo: 1 })
+        StatsD.increment('counter', tags: { foo: 1 })
+      end
+    end
+
+    assert_assertion_triggered do
+      @test_case.assert_statsd_increment('counter', times: 2, tags: ['foo:1']) do
+        StatsD.increment('counter', tags: { foo: 1 })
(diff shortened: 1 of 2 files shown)


test/assertions_test.rb
```diff
@@ -81,6 +81,20 @@ def test_assert_statsd_call
     end
 
     assert_no_assertion_triggered do
+      @test_case.assert_statsd_increment('counter', times: 2, tags: ['foo:1']) do
+        StatsD.increment('counter', tags: { foo: 1 })
+        StatsD.increment('counter', tags: { foo: 1 })
+      end
+    end
+
+    assert_assertion_triggered do
```
fw42: this fails, but I think it should pass

wvanbergen:
:+1:, I think metrics should only be counted if all parameters confirm to the requirements.

test/assertions_test.rb
surrahman: @fw42, I assume you really meant the next test where the `foo` tag has two different values?

fw42: The assertion on this line (90, the one I'm commenting on) is failing, but it should pass. And yea that's the case with two different values.

surrahman:
Quick fix to examine all metric tags .. cc @fw42, @wvanbergen

fw42:
Cool. Change looks good to me. Willem?

wvanbergen:
:+1:, 🚢 🇮🇹 !

PR #81, opened 2016-05-18
Make UDPBackend thread safe

csfrancis:
@Sirupsen 

Replaces https://github.com/Shopify/cookbooks/pull/11048.


Sirupsen:
Great, thanks!

PR #82, opened 2016-05-19
Add type assertion for sample_rate

MalazAlamir:
Adds type assertion for `sample_rate` ensuring its numeric. This is a nice to have, cause I messed up the ordering of a statsd call in a different repo, which caused my hash for `tags` to be interpreted as `sample_rate`. I had tests that didn't catch this, because I didn't add any assertions on the `sample_rate`.  

@wvanbergen @fw42

### DIFF
--- a/lib/statsd/instrument/assertions.rb
+++ b/lib/statsd/instrument/assertions.rb
@@ -42,6 +42,7 @@ module StatsD::Instrument::Assertions
 
     metrics.each do |metric|
 
+      assert within_numeric_range?(metric.sample_rate), "Unexpected sample rate type for metric #{metric_name}, must be numeric"
       assert_equal options[:sample_rate], metric.sample_rate, "Unexpected StatsD sample rate for metric #{metric_name}" if options[:sample_rate]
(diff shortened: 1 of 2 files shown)


lib/statsd/instrument/assertions.rb
```diff
@@ -66,4 +67,8 @@ def assert_statsd_call(metric_type, metric_name, options = {}, &block)
       metric
     end
   end
+
+  def numeric?(object)
+    true if Float(object) rescue false
```
wvanbergen: This is both confusing (name and implementation do not match), and will probably break many tests where we use `sample_rate: 1` 

you can simply do `object.kind_of?(Numeric)`, and if you want a better check, add `(0.0..1.0).cover?(object)`

wvanbergen:
Other than that, I think this is a great addition :+1:.

lib/statsd/instrument/assertions.rb
MalazAlamir: you're right, I wanted to catch numbers wrapped in strings but thats totally unnecessary.

fw42:
Sure, why not 👍

PR #84, opened 2016-05-23
Use rspec-core instead of just rspec for rspec-rails compatibility

### LINKED ISSUE #83: New version breaks with RSpec version ~ 3.4
O hi!

I get an error when I run my tests with version `2.0.12` with `rspec-core` version `3.4.1`:

```
lib/ruby/gems/2.3.0/gems/statsd-instrument-2.0.12/lib/statsd/instrument/matchers.rb:2:in `require': cannot load such file -- rspec/version (LoadError)
```

I looks this was introduced by commit 3dd6033c1e53810b5a3724f3b86a2b2849850f56. Version `2.0.11` seems to work fine.

shaneog:
`rspec-rails` does not include the meta-gem `rspec`, thus requiring
`rspec/version` will fail.
Instead, rely on `rspec-core` which is included by both `rspec` and
`rspec-rails`.

Fixes #83

### DIFF
--- a/lib/statsd/instrument/matchers.rb
+++ b/lib/statsd/instrument/matchers.rb
@@ -1,5 +1,5 @@
 require 'rspec/expectations'
-require 'rspec/version'
+require 'rspec/core/version'
 
 module StatsD::Instrument::Matchers
   CUSTOM_MATCHERS = {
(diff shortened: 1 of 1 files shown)


wvanbergen:
:+1:

knappe:
Ran into this issue today and had to downgrade to 2.0.11 to work around the issue.

PR #86, opened 2016-05-27
Support an ordered set of options when asserting statsd calls

surrahman:
Currently, when we assert metrics recorded from statsd calls, we allow developers to specify a single set of options for metrics. This prevents us from validating cases where we have multiple metrics with the same name and type being recorded, but with different options, e.g., the following:
```
StatsD.increment('counter', tags: { foo: 1 })
StatsD.increment('counter', tags: { foo: 1 })
StatsD.increment('counter', tags: { foo: 2 })
```
This change allows us to specify an ordered array of metric options via `StatsD::Instrument::Assertions#assert_statsd_calls` so the above case could be validated as shown below:

```
foo_1_metric = StatsD::Instrument::MetricExpectation.new(type: :c, name: 'counter', times: 2, tags: ['foo:1'])
foo_2_metric = StatsD::Instrument::MetricExpectation.new(type: :c, name: 'counter', times: 1, tags: ['foo:2'])
@test_case.assert_statsd_calls([foo_1_metric, foo_2_metric]) do
  StatsD.increment('counter', tags: { foo: 1 })
  StatsD.increment('counter', tags: { foo: 1 })
  StatsD.increment('counter', tags: { foo: 2 })
end
```

### DIFF
--- a/lib/statsd/instrument/assertions.rb
+++ b/lib/statsd/instrument/assertions.rb
@@ -31,41 +31,51 @@ module StatsD::Instrument::Assertions
     assert_statsd_call(:kv, metric_name, options, &block)
   end
 
+  # @private
+  def assert_statsd_calls(expected_metrics, &block)
+    metrics = capture_statsd_calls(&block)
+    matched_expected_metrics = []
+
+    expected_metrics.each do |expected_metric|
+      expected_metric_times = expected_metric.times
+      expected_metric_times_remaining = expected_metric.times
+      filtered_metrics = metrics.select { |m| m.type == expected_metric.type && m.name == expected_metric.name }
+      assert filtered_metrics.length > 0,
+        "No StatsD calls for metric #{expected_metric.name} of type #{expected_metric.type} were made."
+
+      filtered_metrics.each do |metric|
+        assert within_numeric_range?(metric.sample_rate),
+          "Unexpected sample rate type for metric #{metric.name}, must be numeric"
+        if expected_metric.matches(metric)
+          assert expected_metric_times_remaining > 0,
+            "Unexpected StatsD call; number of times this metric was expected exceeded: #{expected_metric.inspect}"
+          expected_metric_times_remaining -= 1
+          metrics.delete(metric)
+          if expected_metric_times_remaining == 0
+            matched_expected_metrics << expected_metric
+          end
+        end
+      end
+
+      assert expected_metric_times_remaining == 0,
+          "Metric expected #{expected_metric_times} times but seen"\
+          " #{expected_metric_times-expected_metric_times_remaining}"\
+          " times: #{expected_metric.inspect}"
+    end
+    expected_metrics -= matched_expected_metrics
+
+    assert expected_metrics.empty?,
+      "Unexpected StatsD calls; the following metric expectations were not satisfied: #{expected_metrics.inspect}"
+  end
+
   private
 
   def assert_statsd_call(metric_type, metric_name, options = {}, &block)
+    options[:name] = metric_name
+    options[:type] = metric_type
     options[:times] ||= 1
-    metrics = capture_statsd_calls(&block)
-    metrics = metrics.select { |m| m.type == metric_type && m.name == metric_name }
-    assert metrics.length > 0, "No StatsD calls for metric #{metric_name} were made."
-    assert options[:times] === metrics.length, "The amount of StatsD calls for metric #{metric_name} was unexpected. Expected #{options[:times].inspect}, found #{metrics.length}"
-
-    metrics.each do |metric|
-
-      assert within_numeric_range?(metric.sample_rate), "Unexpected sample rate type for metric #{metric_name}, must be numeric"
-      assert_equal options[:sample_rate], metric.sample_rate, "Unexpected StatsD sample rate for metric #{metric_name}" if options[:sample_rate]
-      assert_equal options[:value], metric.value, "Unexpected value submitted for StatsD metric #{metric_name}" if options[:value]
-
-      if options[:tags]
-
-        expected_tags = Set.new(StatsD::Instrument::Metric.normalize_tags(options[:tags]))
-        actual_tags = Set.new(metric.tags)
-
-        if options[:ignore_tags]
-          ignored_tags = Set.new(StatsD::Instrument::Metric.normalize_tags(options[:ignore_tags])) - expected_tags
-          actual_tags -= ignored_tags
-
-          if options[:ignore_tags].is_a?(Array)
-            actual_tags.delete_if{ |key| options[:ignore_tags].include?(key.split(":").first) }
-          end
-        end
-
-        assert_equal expected_tags, actual_tags,
-                     "Unexpected StatsD tags for metric #{metric_name}. Expected: #{expected_tags.inspect}, actual: #{actual_tags.inspect}"
-      end
-
-      metric
-    end
+    expected_metric = StatsD::Instrument::MetricExpectation.new(options)
+    assert_statsd_calls([expected_metric], &block)
   end
 
   def within_numeric_range?(object)
--- a/lib/statsd/instrument/metric_expectation.rb
+++ b/lib/statsd/instrument/metric_expectation.rb
@@ -1,0 +1,67 @@
+# @private
+class StatsD::Instrument::MetricExpectation
+
+  attr_accessor :times, :type, :name, :value, :sample_rate, :tags
+  attr_reader :ignore_tags
+
+  def initialize(options = {})
+    @type = options[:type] or raise ArgumentError, "Metric :type is required."
+    @name = options[:name] or raise ArgumentError, "Metric :name is required."
+    @name = StatsD.prefix ? "#{StatsD.prefix}.#{@name}" : @name unless options[:no_prefix]
+    @tags = StatsD::Instrument::Metric.normalize_tags(options[:tags])
+    @times = options[:times] or raise ArgumentError, "Metric :times is required."
+    @sample_rate = options[:sample_rate]
+    @value = options[:value]
+    @ignore_tags = StatsD::Instrument::Metric.normalize_tags(options[:ignore_tags])
+  end
+
+  def matches(actual_metric)
+    return false if sample_rate && sample_rate != actual_metric.sample_rate
+    return false if value && value != actual_metric.value
+
+    if tags
+
+      expected_tags = Set.new(tags)
+      actual_tags = Set.new(actual_metric.tags)
+
+      if ignore_tags
+        ignored_tags = Set.new(ignore_tags) - expected_tags
+        actual_tags -= ignored_tags
+
+        if ignore_tags.is_a?(Array)
+          actual_tags.delete_if{ |key| ignore_tags.include?(key.split(":").first) }
+        end
+      end
+
+      return expected_tags.subset?(actual_tags)
+    end
+    true
+  end
+
+  def default_value
+    case type
+      when :c; 1
+    end
+  end
+
+  TYPES = {
+      c:  'increment',
+      ms: 'measure',
+      g:  'gauge',
+      h:  'histogram',
+      kv: 'key/value',
+      s:  'set',
+  }
+
+  def to_s
+    str = "#{TYPES[type]} #{name}:#{value}"
+    str << " @#{sample_rate}" if sample_rate != 1.0
+    str << " " << tags.map { |t| "##{t}"}.join(' ') if tags
+    str << " times:#{times}" if times > 1
+    str
+  end
+
+  def inspect
+    "#<StatsD::Instrument::MetricExpectation #{self.to_s}>"
+  end
+end
--- a/test/assertions_test.rb
+++ b/test/assertions_test.rb
@@ -124,7 +124,7 @@ class AssertionsTest < Minitest::Test
       end
     end
 
-    assert_assertion_triggered do
+    assert_no_assertion_triggered do
       @test_case.assert_statsd_increment('counter', sample_rate: 0.5, tags: { a: 1 }, ignore_tags: { b: 2 }) do
         StatsD.increment('counter', sample_rate: 0.5, tags: { a: 1, b: 3 })
       end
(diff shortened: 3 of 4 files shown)


wvanbergen:
Not sure if I get this. Can you share a code snippet of how this would be used?

wvanbergen:
Also, this really needs tests to ensure it works, and stays working.

surrahman:
Oh wow @wvanbergen, you picked up on this quickly! Just using this to share an idea with @fw42 .. it would need tests for sure .. we're looking at an alternate approach now so this might not be needed at all.

test/assertions_test.rb
```diff
@@ -87,13 +87,43 @@ def test_assert_statsd_call
...
 
+    assert_no_assertion_triggered do
+      foo_1_metric_tags = { times: 2, tags: ['foo:1'] }
+      foo_2_metric_tags = { times: 1, tags: ['foo:2'] }
+      @test_case.assert_statsd_increments('counter', [foo_1_metric_tags, foo_2_metric_tags]) do
+        StatsD.increment('counter', tags: { foo: 1 })
+        StatsD.increment('counter', tags: { foo: 1 })
+        StatsD.increment('counter', tags: { foo: 2 })
+      end
+    end
```
surrahman: @fw42, @wvanbergen, this is the sort of thing I'd like to add support for .. does it make sense to you guys? If there's a better/easier way to do this with what we already have, happy go to that route instead.

lib/statsd/instrument/assertions.rb
```diff
@@ -11,6 +11,10 @@ def assert_statsd_increment(metric_name, options = {}, &block)
     assert_statsd_call(:c, metric_name, options, &block)
   end
 
+  def assert_statsd_increments(metric_name, ordered_metric_options = [], &block)
```
fw42: Not sure if adding a new method is the best choice for the API here... Maybe it's more intuitive to change the old API (and support both the old "syntax" and the new one)?

lib/statsd/instrument/assertions.rb
```diff
@@ -11,6 +11,10 @@ def assert_statsd_increment(metric_name, options = {}, &block)
     assert_statsd_call(:c, metric_name, options, &block)
```
fw42: Not sure, but if we want to go with the new API, then maybe this old method should now just call the new one (with `ordered_metric_options = [ options ]`?)

lib/statsd/instrument/assertions.rb
```diff
@@ -33,39 +37,56 @@ def assert_statsd_key_value(metric_name, options = {}, &block)
...
 
-  def assert_statsd_call(metric_type, metric_name, options = {}, &block)
-    options[:times] ||= 1
+  def assert_statsd_calls(metric_type, metric_name, ordered_metric_options = [], &block)
     metrics = capture_statsd_calls(&block)
     metrics = metrics.select { |m| m.type == metric_type && m.name == metric_name }
     assert metrics.length > 0, "No StatsD calls for metric #{metric_name} were made."
-    assert options[:times] === metrics.length, "The amount of StatsD calls for metric #{metric_name} was unexpected. Expected #{options[:times].inspect}, found #{metrics.length}"
+    expected_calls_count = 0
+    current_metric_index = 0
```
fw42: You can use `each_with_index` instead of `each` and avoid having this variable declared here like this

fw42: actually I'm wrong, was misunderstanding your code. ignore. sorry.

fw42:
I think I understand this better now and I agree that it would be useful (and I don't see how we can achieve something similar without this addition). Code looks good to me too, but I think the PR description needs to be a bit clearer (maybe add an example).

tl;dr 👍 🚢 💨

test/assertions_test.rb
wvanbergen: Is the order important? Or do you just want to assert that all the listed configurations occur at least once? Or exactly once?

test/assertions_test.rb
```diff
@@ -87,13 +87,43 @@ def test_assert_statsd_call
...
+    assert_assertion_triggered do
+      foo_1_metric_tags = { times: 2, tags: ['foo:1'] }
+      foo_2_metric_tags = { times: 1, tags: ['foo:2'] }
+      @test_case.assert_statsd_increments('counter', [foo_1_metric_tags, foo_2_metric_tags]) do
+        StatsD.increment('counter', tags: { foo: 1 })
+        StatsD.increment('counter', tags: { foo: 1 })
+        StatsD.increment('counter', tags: { foo: 2 })
+        StatsD.increment('counter', tags: { foo: 2 })
+      end
+    end
```
wvanbergen: Instead of adding more and more assertions to this test case, I rather add a separate one (or multiple) with descriptions that tell me exactly the behavior that is being tested..

test/assertions_test.rb
surrahman: I think thats a question for whoever is writing the test .. they may want to test that metrics pop up in a certain order .. or they might not care. I haven't seen any tests so far that care about ordering so far. Implementing it with ordering was easy. With respect to number of occurrences, I figured I'd just extend the approach that was already in place .. allow devs to specify the `times` a metric should be seen .. rather than forcing them to repeat asserted metric option sets.

lib/statsd/instrument/assertions.rb
surrahman: I actually started with `each_with_index` .. just because I wanted to use it :p

test/assertions_test.rb
surrahman: Sure, I'll pull these out into another test case called `test_assert_multiple_statsd_calls`.

surrahman: @wvanbergen do you think these tests would suffice?

Amos47:
Ping @wvanbergen. Can we merge this? I could really use it for testing multiple statsD calls with different tags. We can override `times` so it works, but it would be cool to be able to assert the tags without having to add another surrounding block.

wvanbergen:
I think this is a good idea, but I don't really like the implementation. 
- This treats metric names as a special case compared to the other metric attributes
- It feels "bolted on" at the wrong level of abstraction, and doesn't feel like a coherent whole. This makes the API not very consistent and hard to learn. What we need is some kind of expectation object.

I would prefer a (low-level) API like this:

``` ruby
expected_calls = [
  StatsD::MetricExpectation.new(name: 'name', tags: [...])
  StatsD::MetricExpectation.new(name: 'other_name', sample_rate: 1.0)
]

assert_statsd_calls(expected_calls, match: :all / :any) do # default = all
  # ...
end
```

Then all the current assertion methods can be implemented on top of this mechanism.


Does this make sense? Or is it over-engineering?

Amos47:
@wvanbergen I think that makes sense.

The issue right now I think is that we have an assert inside the each we have an assert https://github.com/Shopify/statsd-instrument/pull/86/files#diff-1cdd21b1914856daa678effbadef7645L63. This means we can never assert the second call's tags aside from asserting that it was called with `times`.

fw42:
I think refactoring the implementation makes sense, as long as the external API of the existing assertions doesn't change (we don't want to go through our codebase and change 500 users of those assertion helpers).

surrahman:
Working on refactoring this .. @wvanbergen, is the `match: :all / :any ` really necessary? I think all the tests we have operate under the `all` requirement .. allowing `any` might also result in some metrics disappearing without our noticing (i.e., incorrect usage of the `any` option).

wvanbergen:
I'm not sure. If you can implement the current functionality with just `all`, let's start there. we can always add `any` later.

surrahman:
Cool .. I'll start with `all` and we can add in `any` later, as/when needed.

surrahman:
@wvanbergen, is this closer to what you had in mind?

lib/statsd/instrument/metric_expectation.rb
```diff
@@ -0,0 +1,30 @@
...
+  # @times times [Numeric] :times The number of times this metric is expected
+  # @@ignore_tags @ignore_tags [Array<String>, Hash<String, String>, nil] :@ignore_tags The tags
+  #   to ignore for this metric.
+  def initialize(options = {})
+    super(options)
+    @times = options[:times] or raise ArgumentError, "Metric :times is required."
+    @ignore_tags = StatsD::Instrument::Metric.normalize_tags(options[:ignore_tags])
+  end
+
+  # @private
```
fw42: what is this stuff?

lib/statsd/instrument/metric_expectation.rb
```diff
@@ -0,0 +1,30 @@
+class StatsD::Instrument::MetricExpectation < StatsD::Instrument::Metric
+
+  attr_accessor :times, :ignore_tags
```
fw42: looks like you are never using the writer, so `attr_reader` should be enough?

lib/statsd/instrument/metric_expectation.rb
```diff
@@ -0,0 +1,30 @@
...
+
+  # Initializes a new MetricExpectation instance.
```
fw42: I'd say that's pretty obvious :) we usually don't add boilerplate comments like this

lib/statsd/instrument/metric_expectation.rb
```diff
@@ -0,0 +1,30 @@
...
+  #
+  # @times times [Numeric] :times The number of times this metric is expected
+  # @@ignore_tags @ignore_tags [Array<String>, Hash<String, String>, nil] :@ignore_tags The tags
+  #   to ignore for this metric.
+  def initialize(options = {})
+    super(options)
+    @times = options[:times] or raise ArgumentError, "Metric :times is required."
```
fw42: hm this is conventionally done as `@times = options.fetch(:times)` (which raises if the key doesn't exist)

lib/statsd/instrument/metric_expectation.rb
surrahman: Not needed, will remove

lib/statsd/instrument/metric_expectation.rb
surrahman: Yep!

lib/statsd/instrument/metric_expectation.rb
surrahman: Removed

lib/statsd/instrument/metric_expectation.rb
surrahman: Just following convention in `StatsD::Instrument::Metric#initialize` .. but I can change this.

test/assertions_test.rb
```diff
@@ -149,6 +149,38 @@ def test_assert_statsd_call
     end
   end
 
+  def test_assert_multiple_statsd_calls
+    assert_assertion_triggered do
+      foo_1_metric = StatsD::Instrument::MetricExpectation.new(type: :c, name: 'counter', times: 2, tags: ['foo:1'])
+      foo_2_metric = StatsD::Instrument::MetricExpectation.new(type: :c, name: 'counter', times: 1, tags: ['foo:2'])
+      @test_case.assert_statsd_calls([foo_1_metric, foo_2_metric]) do
+        StatsD.increment('counter', tags: { foo: 1 })
+         StatsD.increment('counter', tags: { foo: 2 })
```
fw42: indentation here is wrong

test/assertions_test.rb
```diff
@@ -149,6 +149,38 @@ def test_assert_statsd_call
...
+      foo_2_metric = StatsD::Instrument::MetricExpectation.new(type: :c, name: 'counter', times: 1, tags: ['foo:2'])
+      @test_case.assert_statsd_calls([foo_1_metric, foo_2_metric]) do
+        StatsD.increment('counter', tags: { foo: 1 })
+        StatsD.increment('counter', tags: { foo: 1 })
+        StatsD.increment('counter', tags: { foo: 2 })
+        StatsD.increment('counter', tags: { foo: 2 })
+      end
+    end
+
+    assert_assertion_triggered do
```
fw42: why does this pass? you are expecting twice foo:1 and once foo:2 and that's what the block does. So shouldn't this be `assert_no_assertion_triggered` then?

test/assertions_test.rb
fw42: I don't think we ever care what order metrics appear in. At least I can't think of any use cases for that. I wouldn't add support for that unless someone explicitly asks for it.

lib/statsd/instrument/assertions.rb
```diff
@@ -31,41 +31,56 @@ def assert_statsd_key_value(metric_name, options = {}, &block)
...
+  end
 
-        expected_tags = Set.new(StatsD::Instrument::Metric.normalize_tags(options[:tags]))
-        actual_tags = Set.new(metric.tags)
+  private
 
-        if options[:ignore_tags]
-          ignored_tags = Set.new(StatsD::Instrument::Metric.normalize_tags(options[:ignore_tags])) - expected_tags
-          actual_tags -= ignored_tags
+  def assert_metric_matches_expectation(actual_metric, expected_metric)
```
fw42: I feel like this method might just as well be an instance method on the expectation class, so you can just say `expected_metric.matches?(actual_metric)` and that returns true or false (and maybe sets some error message or something).

lib/statsd/instrument/assertions.rb
```diff
@@ -31,41 +31,56 @@ def assert_statsd_key_value(metric_name, options = {}, &block)
...
-        end
+    if expected_metric.tags
 
-        assert_equal expected_tags, actual_tags,
-                     "Unexpected StatsD tags for metric #{metric_name}. Expected: #{expected_tags.inspect}, actual: #{actual_tags.inspect}"
+      expected_tags = Set.new(StatsD::Instrument::Metric.normalize_tags(expected_metric.tags))
+      actual_tags = Set.new(actual_metric.tags)
+
+      if expected_metric.ignore_tags
+        ignored_tags = Set.new(StatsD::Instrument::Metric.normalize_tags(expected_metric.ignore_tags)) - expected_tags
```
fw42: `expected_metric.ignore_tags` is already normalized. Why normalize it again?

lib/statsd/instrument/assertions.rb
```diff
@@ -31,41 +31,56 @@ def assert_statsd_key_value(metric_name, options = {}, &block)
...
 
-          if options[:ignore_tags].is_a?(Array)
-            actual_tags.delete_if{ |key| options[:ignore_tags].include?(key.split(":").first) }
-          end
-        end
+    if expected_metric.tags
 
-        assert_equal expected_tags, actual_tags,
-                     "Unexpected StatsD tags for metric #{metric_name}. Expected: #{expected_tags.inspect}, actual: #{actual_tags.inspect}"
+      expected_tags = Set.new(StatsD::Instrument::Metric.normalize_tags(expected_metric.tags))
```
fw42: I think `expected_metric.tags` (i.e. the `tags` instance method) could take care of the normalization and just return already normalized tags.

lib/statsd/instrument/assertions.rb
```diff
@@ -31,41 +31,56 @@ def assert_statsd_key_value(metric_name, options = {}, &block)
...
+      actual_tags = Set.new(actual_metric.tags)
+
+      if expected_metric.ignore_tags
+        ignored_tags = Set.new(StatsD::Instrument::Metric.normalize_tags(expected_metric.ignore_tags)) - expected_tags
+        actual_tags -= ignored_tags
+
+        if expected_metric.ignore_tags.is_a?(Array)
```
fw42: doesn't `normalize_tags` always return an array (I'm not sure)? if so then this will never be false

fw42: actually this is your change. In the original code, this was comparing to the unnormalized list of tags and now you are comparing it to the normalized version. Is that intended? Does that keep the original behaviour?

lib/statsd/instrument/assertions.rb
```diff
@@ -31,41 +31,56 @@ def assert_statsd_key_value(metric_name, options = {}, &block)
...
 
-      metric
+      assert_equal expected_tags, actual_tags,
+        "Unexpected StatsD tags for metric #{actual_metric.name}. Expected: #{expected_tags.inspect}, actual: #{actual_tags.inspect}"
     end
+
+    actual_metric
+  end
+
+  def assert_statsd_call(metric_type, metric_name, options = {}, &block)
```
fw42: Why is `options` a parameter? you are never passing that in and all the code always uses the default of `{}`. I think in that case it shouldn't be a parameter. Why would you want to (or need to) pass in options to the MetricExpectation class? All the possible options are overwritten by you.

lib/statsd/instrument/metric_expectation.rb
```diff
@@ -0,0 +1,30 @@
+class StatsD::Instrument::MetricExpectation < StatsD::Instrument::Metric
+
+  attr_accessor :times, :ignore_tags
+
+  # Initializes a new MetricExpectation instance.
+  #
+  # @times times [Numeric] :times The number of times this metric is expected
+  # @@ignore_tags @ignore_tags [Array<String>, Hash<String, String>, nil] :@ignore_tags The tags
+  #   to ignore for this metric.
+  def initialize(options = {})
```
fw42: Does the rest of this codebase use options hashes? Because if not then I would suggest using keyword arguments instead.

test/assertions_test.rb
surrahman: Yeah that shouldn't pass .. I'll investigate ..

test/assertions_test.rb
surrahman: Will correct

lib/statsd/instrument/assertions.rb
surrahman: Will refector ..

lib/statsd/instrument/assertions.rb
surrahman: Will remove ..

lib/statsd/instrument/assertions.rb
surrahman: Yep .. will change ..

surrahman: Looks like we store tags in "canonical" form (normalized) in the `initialize` method when we create new `StatsD::Instrument::Metric` instances, so maybe we don't need to normalize again?

lib/statsd/instrument/assertions.rb
surrahman: This was the old declaration that I've left unchanged .. I do fill in `name`, `type` and `times`, but not `value` and `sample_rate` or `tags` .. these are then used when comparing recorded metrics against expected metrics in `StatsD::Instrument::MetricExpectation#matches`.

lib/statsd/instrument/metric_expectation.rb
surrahman: Yeah, I just followed the pattern in `StatsD::Instrument::Metric#initialize` ..

test/assertions_test.rb
surrahman: Agreed ..

wvanbergen:
If you can test this branch on Shopify core, I am happy to merge this.

lib/statsd/instrument/metric_expectation.rb
```diff
@@ -29,7 +29,7 @@ def matches(actual_metric)
         end
       end
 
-      return expected_tags == actual_tags
+      return expected_tags.subset?(actual_tags)
```
surrahman: @wvanbergen, I think this is something you should look at. I'm suggesting that the tags we place in our expectation should at least be a subset of the tags that we see in actual metrics generated by our code .. this is based on the usage I've seen in some of the Shopify tests, whereby developers seem to add assertions and only check for the tags they're interested in. What we had here previously was an exact match check .. which is also correct, just more strict. What would you prefer?

test/assertions_test.rb
```diff
@@ -124,7 +124,7 @@ def test_assert_statsd_call
       end
     end
 
-    assert_assertion_triggered do
+    assert_no_assertion_triggered do
```
surrahman: This test change is based on the other change in this commit noted above, i.e. the check for tags is less strict that it used to be in that it'll validate that expected tags are a subset of actual tags for each matching metric.

lib/statsd/instrument/metric_expectation.rb
surrahman: @fw42 does this make sense to you?

test/assertions_test.rb
wvanbergen: I think a subset makes sense. :+1:

If you end up sending more tags, DD always aggregates over them by default anyway. So the presence of additional tags should not break your monitors or dashboard by sending additional tags.

fw42: one argument against it is that you can now no longer assert on the absence of tags.. Like maybe there is a test that says "if a, then this tag is present, and if not a, then this tag is not present"...

wvanbergen: I think the simple syntax should cater for the most common use case.

If we want to support more advanced use cases, we can make the `MetricExpectation` class have a more advanced "matcher" API (e.g. for assertion that a tag is absent). Users can construct instances an send them to `assert_statsd_calls`. Let's only cross that bridge when there's a need for it though.

surrahman: 👍 I've marked it as `# @private`, as discussed on Slack.

wvanbergen:
We've opted to tag `MetricExpectation`, and `assert_statsd_calls` as private, because I am not ready yet to commit to not breaking the API of those for now.

Goldenson:
## Note
FYI, this PR breaks the implementation of `assert_statsd_increment` when dealing with `StatsD.prefix`.

## Example

Setting `StatsD.prefix = 'WhateverPrefix'`

🔴  This is not working anymore:
```
assert_statsd_increment('WhateverPrefix.message') do
  ...
end
```

✅ This is working:
```
assert_statsd_increment('message') do
  ...
end
```

Some folks fixed their tests by removing the prefix in this [PR](https://github.com/Shopify/remember-me/pull/107).

I'm not sure what's the expected behaviour here, just want to report it.

cc @christhomson @wvanbergen

PR #87, opened 2016-06-15
Support a value keyword argument for all metric types.

wvanbergen:
This will update the library to support a `value` key in the option hash for all metric calls. This logic can be cleaned up after we move towards keyword arguments.

@fw42 @surrahman

### DIFF
--- a/test/statsd_test.rb
+++ b/test/statsd_test.rb
@@ -17,12 +17,25 @@ class StatsDTest < Minitest::Test
     assert_equal :ms, metric.type
   end
 
+  def test_statsd_measure_with_explicit_value_as_keyword_argument
+    result = nil
+    metric = capture_statsd_call { result = StatsD.measure('values.foobar', value: 42) }
+    assert_equal metric, result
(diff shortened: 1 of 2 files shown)


test/statsd_test.rb
```diff
@@ -17,12 +17,25 @@ def test_statsd_measure_with_explicit_value
...
+    result = nil
+    metric = capture_statsd_call { result = StatsD.measure('values.foobar', value: 42) }
+    assert_equal metric, result
+    assert_equal 'values.foobar', metric.name
+    assert_equal 42, metric.value
+    assert_equal :ms, metric.type
+  end
+
+  def test_statsd_measure_without_value_or_block
+    assert_raises(ArgumentError) { StatsD.measure('values.foobar', tags: 123) }
```
fw42: I thought we default to a value of 1?

fw42: I guess only for `increment`

fw42:
👍 💃

PR #90, opened 2016-08-02
Ask Rails for the environment when Rails is loaded and not just defined

reprah:
I'm working on an app that uses statsd-instrument alongside other libraries. One of those libraries defines `Rails`, but that version of Rails does not have `env` defined on it. This raises an exception when the app boots:

`.../.rbenv/versions/2.3.1/lib/ruby/gems/2.3.0/gems/statsd-instrument-2.1.1/lib/statsd/instrument/environment.rb:32:in 'environment': undefined method 'env' for Rails:Module (NoMethodError)`

This happens because statsd-instrument's environment detecting method checks if Rails is defined, and if it is, calls `env` on it. This PR adds an additional guard and a test.

@jstorimer I believe I'm supposed to ping you for a code review, please? Let me know if you have suggestions or know of a reason this change is inappropriate. Thanks!

### DIFF
--- a/test/environment_test.rb
+++ b/test/environment_test.rb
@@ -1,5 +1,7 @@
 require 'test_helper'
 
+module Rails; end
+
 class EnvironmentTest < Minitest::Test
 
   def setup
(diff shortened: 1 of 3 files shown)


test/environment_test.rb
```diff
@@ -32,4 +34,8 @@ def test_uses_environment_variables_in_production_environment
     assert_equal 1234, backend.port
     assert_equal :datadog, backend.implementation
   end
+
+  def test_uses_env_when_rails_does_not_respond_to_env
+    assert_equal ENV['ENV'], StatsD::Instrument::Environment.environment
+  end
```
wvanbergen: Can you also add a test that ensures everything works as expected if `Rails.env` is available?

wvanbergen:
Minor comment. I will merge once you add the test. Thanks!

reprah:
@wvanbergen Thanks, I added the test.

wvanbergen:
Thanks!

PR #92, opened 2016-08-15
Use prepended modules instead of modifying the target class

### LINKED ISSUE #91: Calling `statsd_measure` after prepending a module causes infinite loop
To reproduce:

```ruby
require "bundler/inline"

gemfile(true) do
  gem "statsd-instrument", github: "shopify/statsd-instrument"
end

class Foo
  def foo
    puts "foo"
  end
end

module MyPatch
  def foo
    puts "before foo"
    super
  end
end

Foo.prepend(MyPatch)

Foo.extend(StatsD::Instrument)
Foo.statsd_measure :foo, "Foo!"

Foo.new.foo
```

Output:

```
before foo
before foo
before foo
before foo
before foo
before foo
before foo
before foo
before foo
/Users/sean/.rbenv/versions/2.3.1/lib/ruby/gems/2.3.0/bundler/gems/statsd-instrument-154487eb6a1d/lib/statsd/instrument.rb:75:in `block (3 levels) in statsd_measure': stack level too deep (SystemStackError)
	from /Users/sean/.rbenv/versions/2.3.1/lib/ruby/gems/2.3.0/bundler/gems/statsd-instrument-154487eb6a1d/lib/statsd/instrument.rb:284:in `block in measure'
	from /Users/sean/.rbenv/versions/2.3.1/lib/ruby/gems/2.3.0/bundler/gems/statsd-instrument-154487eb6a1d/lib/statsd/instrument.rb:53:in `duration'
	from /Users/sean/.rbenv/versions/2.3.1/lib/ruby/gems/2.3.0/bundler/gems/statsd-instrument-154487eb6a1d/lib/statsd/instrument.rb:284:in `measure'
	from /Users/sean/.rbenv/versions/2.3.1/lib/ruby/gems/2.3.0/bundler/gems/statsd-instrument-154487eb6a1d/lib/statsd/instrument.rb:75:in `block (2 levels) in statsd_measure'
	from foo.rb:16:in `foo'
	from /Users/sean/.rbenv/versions/2.3.1/lib/ruby/gems/2.3.0/bundler/gems/statsd-instrument-154487eb6a1d/lib/statsd/instrument.rb:75:in `block (3 levels) in statsd_measure'
	from /Users/sean/.rbenv/versions/2.3.1/lib/ruby/gems/2.3.0/bundler/gems/statsd-instrument-154487eb6a1d/lib/statsd/instrument.rb:284:in `block in measure'
	from /Users/sean/.rbenv/versions/2.3.1/lib/ruby/gems/2.3.0/bundler/gems/statsd-instrument-154487eb6a1d/lib/statsd/instrument.rb:53:in `duration'
	 ... 10758 levels...
	from /Users/sean/.rbenv/versions/2.3.1/lib/ruby/gems/2.3.0/bundler/gems/statsd-instrument-154487eb6a1d/lib/statsd/instrument.rb:284:in `measure'
	from /Users/sean/.rbenv/versions/2.3.1/lib/ruby/gems/2.3.0/bundler/gems/statsd-instrument-154487eb6a1d/lib/statsd/instrument.rb:75:in `block (2 levels) in statsd_measure'
	from foo.rb:16:in `foo'
	from foo.rb:25:in `<main>'
```

This would likely be solved by changing `statsd_measure` to use `Module.prepend`, which this gem can now do as it doesn't support 1.9.3

sgrif:
If the method being overidden by this gem is also overidden by a prepended
module, the result ends up being infinite recursion as `old_method` will always
be pointing to the method added by the module, and `super` there will point to
the addition from `StatsD::Instrument`. Normally I would just prepend a single
anonymous module every time this was called, but I've attempted to preserve the
"has this already been instrumented" checks and the ability to remove
instrumentations later. Instead we keep a hash of various modules, and
define/undefine the method being overidden. It's impossible to remove a module
after it's been added, but when `statsd_remove_measure` or friends are called
we simply undefine the methdos on the module leaving us with an empty module
doing nothing.

This does mean that we'll end up with a lot of anonymous modules in the
ancestors chain, but there's really no way to avoid that. I have given them a
more descriptive name at least in order to aid debugging if that occurs

Fixes #91.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,7 +3,6 @@ rvm:
   - 2.0
   - 2.1
   - 2.2
-  - jruby-19mode
   - rbx-2
   - ruby-head
 
--- a/lib/statsd/instrument.rb
+++ b/lib/statsd/instrument.rb
@@ -44,7 +44,7 @@ module StatsD
     def statsd_instrumentations
       if defined?(@statsd_instrumentations)
         @statsd_instrumentations
-      elsif superclass.respond_to?(:statsd_instrumentations)
+      elsif respond_to?(:superclass) && superclass.respond_to?(:statsd_instrumentations)
         superclass.statsd_instrumentations
       else
         @statsd_instrumentations = {}
--- a/lib/statsd/instrument.rb
+++ b/lib/statsd/instrument.rb
@@ -40,6 +40,16 @@ module StatsD
   # StatsD metrics. E.g., yopu can create counters on how often a method is called, how often it is
   # successful, the duration of the methods call, etc.
   module Instrument
+    # @private
+    def statsd_instrumentations
+      if defined?(@statsd_instrumentations)
+        @statsd_instrumentations
+      elsif superclass.respond_to?(:statsd_instrumentations)
+        superclass.statsd_instrumentations
+      else
+        @statsd_instrumentations = {}
+      end
+    end
 
     # @private
     def self.generate_metric_name(metric_name, callee, *args)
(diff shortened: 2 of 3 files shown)


sgrif:
Review? @wvanbergen

wvanbergen:
It creates only one anonymous module per class/module, right? Not one for every method being instrumented? That's the best we can do I think.

sgrif:
No it has to be one per instrumentation per method, since the user can add multiple instrumentations to the same method

sgrif:
@wvanbergen This look good to merge?

wvanbergen:
:+1:

PR #93, opened 2016-09-16
Adds filter to capture_statsd_calls

pallan:
The default `capture_statsd_calls` behaviour is to, naturally, capture everything. On occasion you may have other StatsD calls in the execution flow that are not relevant to your test and appear in the returned metrics. These irrelevant metrics need to be filtered out in order to perform you assertions.

This PR adds an optional `filter` argument to the command that will only return metrics whose name matches the filter. The argument accepts either a String or Regexp as valid match. 

@wvanbergen @fw42

### DIFF
--- a/lib/statsd/instrument/helpers.rb
+++ b/lib/statsd/instrument/helpers.rb
@@ -1,9 +1,14 @@
 module StatsD::Instrument::Helpers
-  def capture_statsd_calls(&block)
+  def capture_statsd_calls(filter: nil, &block)
     mock_backend = StatsD::Instrument::Backends::CaptureBackend.new
     old_backend, StatsD.backend = StatsD.backend, mock_backend
     block.call
-    mock_backend.collected_metrics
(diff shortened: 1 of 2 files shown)


wvanbergen:
When I wrote this, I explicitly decided not to add this because it's a leaky abstraction. With this change, you can do a positive filter for metric names using a regular expression. But what about negative filters? What about tags?

I prefer just letting users using the Metric objects that are returned. What is so bad about this?

``` ruby
metrics = capture_statsd_calls { ... }
metrics.reject! { |metric| metric.name =~ /regexp/ }
```

pallan:
Its not that bad :-)  That is how we solved our problem. We can keep it that way if that was part of the design plan.

wvanbergen:
You may be able to convince me otherwise. An alternative could be adding a `filter_statsd_metrics` method that just does the filtering, and not capturing, if you want to make the common case simpler.

PR #95, opened 2016-11-24
Topic/rspec matchers improvements

codegourmet:
content of pull request (all RSpec Matchers related):

1) fixed bug where :times combined with additional expectations would always succeed even if values were different (added tests as proof: _with_times_and_value_matched/not_matched)

2) actual vs. expected failure message output

3) support for rspec-mock style argument matchers (added tests as proof: _argument_matcher_matched/not_atched)

explanation:

1) see test and first commit message (448f3af9a610edfe23222345c3741c34e23e60a8)

2) the matcher captures all statsd calls, so they won't show up in the statsd log (if using logger backend), making debugging difficult. the usual rspec style actual/expected comparison solves this. 

3) vitally important for the use case 'measurement': for example, Timecop isn't entirely precise, resulting in a measurement of 0.5001 if calling Timecop.travel(0.5). rspec argument matchers solve this by not being restricted to :== comparing floats. See example in README.md.

guidelines for conrtibuting:

(ping as requested: @jstorimer @wvanbergen)

I tried to match your coding style, but there was no rubocop config, so if you're concerned about any flavor I introduced, feel free to edit yourselves or request a change, I'm happy to oblige!

The Travis build didn't pass for 'rbx-2', but that was already failing in master.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@ please at an entry to the "unreleased changes" section below.
 - Add support for Ruby 2.2.
 - Make library compatible with RSpec 2.
 - Ask Rails for the environment if Rails is loaded, not just defined
+- RSpec: verbose failure messages (expected vs. actual), rspec-mock argument matchers support
 
 ### Version 2.0.6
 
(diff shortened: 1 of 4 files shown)


wvanbergen:
Awesome, thanks for the contribution!

PR #100, opened 2017-03-23
Friendly assertion error if tags do not match

kirs:
In case your code is:

```ruby
StatsD.increment('counter', tags: { tag1: "MyJob" })
```

and your expectation is:

```ruby
assert_statsd_increment('counter', tags: { tag1: "AnotherJob" }) do
```

The assertion message should be friendly and give a tip that there was another event with the same key. Something similar to what Webmock does.

review @wvanbergen


wvanbergen:
I've rebased this PR.

PR #101, opened 2017-03-24
Added StoreBackend - initial version

lzap:
This might be weird at first sight, but my motivation is to allow collecting statsd data in binary files and them compressing them and sending them over for inspection. The implementation is as simple as possible - the binary format has two types of records, one is metric name and type, the other is value. The main idea is to store numeric ID and type for each individual metric, serialize it and also store in an array so it can be reused. The same process for reading.

The provider accepts either IO or directory and in the latter case it creates subdirectory for every day and re-creates new every midnight so they can be easily rotated, compressed or analyzed on a daily basis. Each process (PID) creates new file, all files are opened append only. The implementation is thread safe.

I can provide documentation, example parser that turns the binary data into STATSD protocol (can be easily set via TCP into server) and provide more changes. There are currently some limitations:

* maximum metric name length is 256 chars (or error is raised)
* type can't change for one archive file (or error is raised)
* only C, MS, G a H types are supported (or error)
* numbers are stored as signed int 64 and floats as regular float

TODO list:

- [ ] use array instead of hash for metric definitions
- [x] name id should be short instead int
- [x] store names as UTF8 encoded with size instead fixed 256
- [ ] consider using 32 bit numeric types
- [ ] write conversion tool
- [ ] write documentation

Interested in what ya think! Cheers.


lzap:
I will change the fixed length metric name for length + UTF8 encoded name which will be nicer. Not yet sure if to go with 32bit or 64bit integers (and double or float), there is no spec, but I don't want to waste space. Let me know what you think, this is not ready to be merged yet.

wvanbergen:
Couple of questions about your approach:

- What are the advantages over the specialized binary format rather than using the StatsD format, compressed?
- Have you considered dumping the UDP traffic to file? E.g. run `nc -u -l 8125 | gzip > out.file` to get a UDP server that simply echoes everything into a file. 
- Have you done any benchmarking? The nice part of instrumenting stuff with UDP is that the overhead is minimal. Writing to disk may be significantly more heavy, making it less suitable for production environments. Not sure though, which is why it could be nice to see some comparisons.

lib/statsd/instrument/backends/store_backend.rb
```diff
@@ -0,0 +1,73 @@
...
+      @names = {}
+      @next_id = 1
+    end
+
+    def collect_metric(m)
+      if m.sample_rate < 1.0 && rand > m.sample_rate
+        return false
+      end
+
+      reopen if @next_reopen && Time.now.to_i > @next_reopen
```
wvanbergen: Rather than dealing with rotating log files here, I'd prefer to use a standard unix like `logrotate` to do that for us. That would simplify the code here significantly, and logrotate probably does a better job.

lib/statsd/instrument/backends/store_backend.rb
```diff
@@ -0,0 +1,73 @@
...
+      ms: 'Qf',
+      g:  'Qf',
+      h:  'Qf',
+    }
+    MetricNameStruct = Struct.new(:id, :name, :type) do
+      def to_bin
+        raise "Metric name cannot be longer than #{MAXIMUM_NAME_LENGTH}" if name.size > MAXIMUM_NAME_LENGTH
+        [id, name, type.to_s].pack("QA#{MAXIMUM_NAME_LENGTH}A2")
+      end
+    end
```
wvanbergen: I assume you have the separate struct to save some space. However, if you're compress this data, it will do something like this for you already.

lib/statsd/instrument/backends/store_backend.rb
```diff
@@ -0,0 +1,73 @@
...
+    MetricStruct = Struct.new(:id, :value, :type) do
```
wvanbergen: You definitely want to include `sample_rate` here (at least when it is not equal to 1). This allows you to know that we suppressed a bunch of events when analyzing the data.

Personally, I'd also like it if tags would the included as well.

test/store_backend_test.rb
```diff
@@ -0,0 +1,65 @@
...
+    fb.collect_metric(StatsD::Instrument::Metric::new(type: :c, name: 'one.c1'))
+    fb.collect_metric(StatsD::Instrument::Metric::new(type: :ms, name: 'mock.ms', value: 1234))
+    fb.collect_metric(StatsD::Instrument::Metric::new(type: :g, name: 'mock.g', value: 56789))
+    file = File.join(tmp_dir, Time.now.strftime("%Y-%m-%d"), "#{$$}.statsb")
+    assert File.exist?(file)
+    fb.reopen
+    assert_equal 1160, File.size(file)
+  ensure
+    FileUtils.rm_rf(tmp_dir) if tmp_dir.start_with?('/tmp')
+  end
```
wvanbergen: Rather than the ensure block, I'd prefer 

``` ruby
Dir.mktmpdir do |tmp_dir|
   ...
end
```

lzap:
> What are the advantages over the specialized binary format rather than using the StatsD format, compressed?

It's so much more efficient. Let's say average metric name length is 30 characeters, then each individual measurement in minimally 34 bytes while in my current design it's just 16bit + 64bit = 6 bytes (well the first measurement is actually the name and id and type which is name length plus few bytes but this is irrelevant).

> Have you considered dumping the UDP traffic to file? E.g. run nc -u -l 8125 | gzip > out.file to get a UDP server that simply echoes everything into a file. This way you could use the standard UDP backend. It could be nice if this gem would include a parser for the StatsD format to help with analysis in that case.

Sure, it's very inefficient. It's just the text based statsd protocol.

> Have you done any benchmarking? The nice part of instrumenting stuff with UDP is that the overhead is minimal. Writing to disk may be significantly more heavy, making it less suitable for production environments. Not sure though, which is why it could be nice to see some comparisons.

Not yet but I plan to indeed. It's a valid point, but performance is not the issue in my case. We have a large userbase of our Rails app deployed on their sites and we want an easy way to get some telemetry data "on demand". It is not supposed to be running all the time, for this we will indeed recommend UDP writer which is more suitable for this.

My idea is to ship the application configured with this new provider (store on disk) and allow our users to reconfigure to UDP for long-term monitoring.

My patch is rough idea, please do not do formal review yet there are some major flaws in there (it can't be deserialized correctly yet), I am working on it. Thanks for your opinion.

lib/statsd/instrument/backends/store_backend.rb
lzap: Yeah this was just my friday idea, I was hacking very quickly. I will drop the 256 name limit and simply store string with its length which is without limitations.

lib/statsd/instrument/backends/store_backend.rb
lzap: Currently measurement is only 6 bytes and I want to keep on that, but I think I know how to store sample rate in separate record type (it does not change often right?) so it will only be stored when it's changed.

lib/statsd/instrument/backends/store_backend.rb
lzap: I considered that and decided to go my own path. If you want logrotate, you still need to implement reopen anyway (usually via SIGHUP or SIGUSR1) which is as complex as this. And this is not log file but rather an "archive" with data. Also I want to "split" archives exactly at midnight, logrotate usually can't do that unless you heavily reconfigure it.

test/store_backend_test.rb
lzap: Absolutely, good idea.

wvanbergen:
> > What are the advantages over the specialized binary format rather than using the StatsD format, compressed?
>
> It's so much more efficient. Let's say average metric name length is 30 characeters, then each individual measurement via statsd protocol (e.g. UDP backend) in minimally 34 bytes while in my current design it's just 16bit + 64bit = 6 bytes (well the first measurement is actually the name and id and type which is name length plus few bytes but this is irrelevant).

> > Have you considered dumping the UDP traffic to file? E.g. run nc -u -l 8125 | gzip > out.file to get a UDP server that simply echoes everything into a file. This way you could use the standard UDP backend. It could be nice if this gem would include a parser for the StatsD format to help with analysis in that case.
>
> Sure, it's very inefficient. It's just the text based statsd protocol.

For both of these, you seemed to have missed my "compressed" mention. Yes, the StatsD protocol is not super space efficient, but the compression should take care of most of the repetition and get you in the same ballpark as the binary format. 

With a custom designed binary format you can probably be slightly more space efficient than compressed (gzip/bzip2) raw StatsD data, but you also give up a lot by going down this path.
- It's not human readable.
- You need extra software to read/interpret the data; none of the standard StatsD tooling will work for you.

I wouldn't want to give these up, unless I absolutely needed the extra space efficiency.

> We have a large userbase of our Rails app deployed on their sites and we want an easy way to get some telemetry data "on demand". It is not supposed to be running all the time, for this we will indeed recommend UDP writer which is more suitable for this.
> 
> My idea is to ship the application configured with this new provider (store on disk) and allow our users to reconfigure to UDP for long-term monitoring.

I wonder if this may be too specific of a use case and whether it makes sense to have in this library as a result. Because backends are pluggable, you can easily load this backend from elsewhere (private code, or your own gem) and activate it. Maybe I am missing something though.

lzap:
Indeed, the efficiency is the primary driver for me. To be honest I hacked with JSONB and MESSAGEPACK protocols/formats but I ended up in coding own one as I want it to be ultra-efficient and this is very easy case - no rocket science here.

It's binary and own format, I agree, but I wanted to write a binary that would convert it from statsb to statsd (as a script in bin/ folder) so users of the library could do the stuff. Just to explain my use case, I will also write my own convertor to load the data into PCP analysis framework for further analysis, but that's out of scope of this PR.

> I wonder if this may be too specific of a use case and whether it makes sense to have in this library as a result. Because backends are pluggable, you can easily load this backend from elsewhere (private code, or your own gem) and activate it. Maybe I am missing something though.

Sure, it's a bit specific I thought it would be useful. That's why I filed it here, but I kinda expected that and I will put it into my own application. How about this, you are not interested in this and you think it does not belong here, just close the PR. If you want to keep an eye on my work and reconsider it later, keep it opened and I will push my final version so you can decide.

wvanbergen:
I am happy to keep this PR open for now as you solidify your ideas, and will review the code for you if you want to, but I think this feature is not really a great fit for this library.

The StatsD protocol is nice because is simple and straightforward, and a community around it exists for client and server applications that share the same protocol. The intention for this library is mainly to be a client that can generate StatsD packets, and you're incredible flexible with what you can do with them by simply listening on the UDP port and process the data as you wish.

If I were to solve your problem, I rather would write a simple StatsD server accepts the UDP packets that writes the data to files like you're doing here, and run it as a small daemon on your servers, so the UDP traffic is all loopback, so very efficient.

lzap:
So here is my almost final version, working now. Few statistics from my dev setup and the difference is big:

```
-rw-rw-r-- 1 lzap lzap  35K Mar 26 22:49 14056.statsb
-rw-rw-r-- 1 lzap lzap 170K Mar 26 22:56 14056.statsd
```

Compressing is indeed pretty effective.

```
-rw-rw-r-- 1 lzap lzap 9.0K Mar 26 22:49 14056.statsb.bz2
-rw-rw-r-- 1 lzap lzap 9.8K Mar 26 22:56 14056.statsd.bz2
```

Anyway, nothing I can do. Also I need to store time as well. I will just push for the record ;-) Take care.

PR #102, opened 2017-05-16
Allow DNS retry on SocketErrors

stonith:
Currently, when using a DNS hostname for the statsd endpoint and a SocketError is rescued, a retry DNS lookup is never made and statsd-instrument never recovers. By invalidating the socket object on error, a DNS lookup is made when the object is recreated allowing statsd-instrument to recover. @csfrancis I'm not sure if this might cause a DoS on non DNS related errors?


wvanbergen:
Is it possible to add a test for this?

stonith:
@csfrancis You're right, this is not a statsd_instrument problem, fluentd where I was seeing this problem sets the `resolve_name` option to false for UDP sockets.

PR #103, opened 2017-06-15
Assertions don't fail on exceptions

fw42:
Imagine a test like this:
```ruby
test "foo" do
      raise "hello"
    end
  end
end
```
This test passes, even though no StatsD metric was emitted. This is super confusing.

I'm not sure what the best way to fix this is, but I think the test in this PR captures the problem.

Any thoughts? @wvanbergen

### DIFF
--- a/test/assertions_test.rb
+++ b/test/assertions_test.rb
@@ -149,6 +149,25 @@ class AssertionsTest < Minitest::Test
     end
   end
 
+  def test_assert_calls_while_exception
+    assert_no_assertion_triggered do
+      assert_raises(RuntimeError) do
+        @test_case.assert_statsd_increment('counter') do
+          StatsD.increment('counter')
+          raise "foo"
+        end
+      end
+    end
+
+    assert_assertion_triggered do
+      assert_raises(RuntimeError) do
+        @test_case.assert_statsd_increment('counter') do
+          raise "foo"
+        end
+      end
+    end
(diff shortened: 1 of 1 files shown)


test/assertions_test.rb
```diff
@@ -149,6 +149,25 @@ def test_assert_statsd_call
...
+    assert_no_assertion_triggered do
+      assert_raises(RuntimeError) do
+        @test_case.assert_statsd_increment('counter') do
+          StatsD.increment('counter')
+          raise "foo"
+        end
+      end
+    end
+
+    assert_assertion_triggered do
```
fw42: This currently fails

wvanbergen: I assume this works as expected, right?
``` ruby
@test_case.assert_statsd_increment('counter') do
    raise "foo"
  end
end
```

fw42: yes, if you turn the order around, then it works

wvanbergen:
This is definitely counter-intuitive at least.

I am happy to see a fix for this but it's not super straightforward for sure. If we fail an assertion, we will end up raising a Minitest::Assertion exception, which might throw off the `assert_raises` that wraps it.

fw42:
@kirs any idea what a good fix could look like?

wvanbergen:
#166 will fix this.

PR #105, opened 2017-07-06
Handle ThreadError on exit signal

andremedeiros:
This is a different approach to tackling the thread safety issue we've had earlier.

Instead of removing it altogether, we rescue ThreadError which will be raised by `synchronize` when the program is exiting, and attempt to send the stat without synchronization.

cc. @fw42 @csfrancis @Shopify/observability

### DIFF
--- a/lib/statsd/instrument/backends/udp_backend.rb
+++ b/lib/statsd/instrument/backends/udp_backend.rb
@@ -64,7 +64,7 @@ module StatsD::Instrument::Backends
     def generate_packet(metric)
       command = "#{metric.name}:#{metric.value}|#{metric.type}"
       command << "|@#{metric.sample_rate}" if metric.sample_rate < 1 || (implementation == :statsite && metric.sample_rate > 1)
-      if metric.tags 
+      if metric.tags
         if tags_supported?
           command << "|##{metric.tags.join(',')}"
         else
@@ -84,6 +84,11 @@ module StatsD::Instrument::Backends
       synchronize do
         socket.send(command, 0) > 0
       end
+    rescue ThreadError => e
+      # In cases where a TERM or KILL signal has been sent, and we send stats as
+      # part of a signal handler, locks cannot be acquired, so we do our best
+      # to try and send the command without a lock.
+      socket.send(command, 0) > 0
     rescue SocketError, IOError, SystemCallError => e
       StatsD.logger.error "[StatsD] #{e.class.name}: #{e.message}"
     end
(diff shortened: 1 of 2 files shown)


test/udp_backend_test.rb
```diff
@@ -132,4 +132,16 @@ def test_io_error_should_not_raise_but_log
...
+  def test_thread_error_does_not_raise
+    pid = fork do
+      Signal.trap('TERM') do
+        StatsD.increment('exiting')
+      end
+
+      sleep 100
+    end
+
+    Process.kill('TERM', pid)
```
csfrancis: Is it possible to assert the statsd request was sent?

test/udp_backend_test.rb
```diff
@@ -132,4 +132,16 @@ def test_io_error_should_not_raise_but_log
     @logger.expects(:error)
     StatsD.increment('fail')
   end
+
+  def test_thread_error_does_not_raise
```
fw42: not sure I understand this test... shouldn't there be some kind of test assertion? or is this basically just testing that nothing raises? what did it raise before?

test/udp_backend_test.rb
```diff
@@ -132,4 +132,16 @@ def test_io_error_should_not_raise_but_log
...
+    pid = fork do
+      Signal.trap('TERM') do
+        StatsD.increment('exiting')
```
fw42: shouldn't we assert that this metric is actually being emitted?

test/udp_backend_test.rb
andremedeiros: The thing is that it might, or might not be sent. I don't think we have any guarantees. Let me 👀

test/udp_backend_test.rb
andremedeiros: It's a test that nothing raises (and exits cleanly.) Acquiring a lock when exiting raises a `ThreadError`, so that's what I'm rescuing.

lib/statsd/instrument/backends/udp_backend.rb
```diff
@@ -84,6 +84,11 @@ def write_packet(command)
       synchronize do
         socket.send(command, 0) > 0
       end
+    rescue ThreadError => e
```
kirs: I'm just curious, shouldn't it be re-raised after we do the socket magic?

andremedeiros: It shouldn't, and here's why: if we re-raise in the case of a container sending a stat on exit, it will cause it to shut down in an unclean way, which makes deploys fail.

Issue #110, opened 2017-10-30
Some metrics seems to be skipped on exceptions


tjoyal:
Based of some dashboard data it lead me to believe we do not properly collect the data correctly on exceptions (at least `measure`). I didn't explore it with more depth but looking at the code that seems to be the case (should at least add a test to cover that).

Eg.: https://github.com/Shopify/statsd-instrument/blob/46df2c67ec9a9384b9b667df13497a884d560278/lib/statsd/instrument.rb#L85

A good scenario where this would occur often by design is when measuring an http request that have an open or read timeout. In these cases, we wouldn't not record failures and it would lead the event consumers to believe "it never happen" while in reality we are simply not issuing event for them.

cc @maximebedard

maximebedard:
I've been able to reproduce fairly easily. I'll see if I can put up a fix ASAP

```rb
irb(main):021:0> StatsD.backend = StatsD::Instrument::Backends::LoggerBackend.new(Logger.new(STDOUT))
irb(main):019:0> StatsD.measure('werwerwer'){puts "asdasd"}
asdasd
I, [2017-10-30T13:05:01.074645 #96755]  INFO -- : [StatsD] measure werwerwer:0.023999949917197227
=> nil
irb(main):020:0> StatsD.measure('werwerwer'){raise "asdasd"}
RuntimeError: asdasd
	from (irb):20:in `block in irb_binding'
	from /Users/maximebedard/.gem/ruby/2.3.3/gems/statsd-instrument-2.1.4/lib/statsd/instrument.rb:287:in `block in measure'
	from /Users/maximebedard/.gem/ruby/2.3.3/gems/statsd-instrument-2.1.4/lib/statsd/instrument.rb:63:in `duration'
	from /Users/maximebedard/.gem/ruby/2.3.3/gems/statsd-instrument-2.1.4/lib/statsd/instrument.rb:287:in `measure'
	from (irb):20
	from /opt/rubies/2.3.3/bin/irb:11:in `<main>'
```

PR #111, opened 2017-10-30
Send measure metric even when the block raises

maximebedard:
Fixes https://github.com/Shopify/statsd-instrument/issues/110

In order to fix this issue, I had to break down the duration method since we need to handle situation where the block would raise in order to calculate the remaining time. Since it was the only place used, I decided to expose the `gettime` method that uses the method available to gather the current time.


lib/statsd/instrument.rb
```diff
@@ -58,17 +58,13 @@ def self.generate_metric_name(metric_name, callee, *args)
 
     if Process.respond_to?(:clock_gettime)
       # @private
-      def self.duration
-        start = Process.clock_gettime(Process::CLOCK_MONOTONIC)
-        yield
-        Process.clock_gettime(Process::CLOCK_MONOTONIC) - start
+      def self.gettime
```
wvanbergen: This method is part of the public API. We cannot really change it like this. I know we have code out there that uses this function

lib/statsd/instrument.rb
```diff
@@ -277,17 +273,25 @@ def backend
...
+      begin
+        result = yield
+      ensure
+        value = 1000 * StatsD::Instrument.gettime - start
+        collect_metric(hash_argument(metric_options).merge(type: :ms, name: key, value: value))
+        result
+      end
+    else
+      collect_metric(hash_argument(metric_options).merge(type: :ms, name: key, value: value))
+    end
```
wvanbergen: I think you can still use the old duration method. 

``` ruby
value = 1000 * StatsD::Instrument.duration do
  begin
    result = yield
  ensure
    ...
  end
end
```

maximebedard: The problem is that we need to have the duration by the time we reach the `ensure` block in order to deliver the metric with the actual length of the call. I dont think this would work :(

maximebedard:
> most likely have much different characteristics compared to calls that are successful.

True, I haven't considered it that way. Should we publish a different metrics in this case?

maximebedard:
TBH I'm not 100% sure what the correct behaviour is. For compairaison, datadog's library also delivers the metric when the block raises https://github.com/DataDog/dogstatsd-ruby/blob/master/lib/datadog/statsd.rb#L203-L204. 

For me, I believe the noise may be a good thing since it will skew metrics and potentially help pinpoint errors/unexpected behaviour.

tjoyal:
> One issue is that calls that fail with an exception most likely have much different characteristics compared to calls that are successful. This simply combines those two histograms into one.

I see your concern and to tell the truth I might be biased by the recent work I've been doing. 

The way I see it I want this gem to "tell me information on what happens when I run this block" and silencing it is counter intuitive for me. (eg.: measuring around an http call might tell me all is fine but in reality a lot of requests timeout)

Something else to note, other events would still track failures. eg.:
`statsd_count_success` would keep track of exceptions
`statsd_count` increment before yielding the old method that might fail

`statsd_count_if` would on the other end not consider failures

So maybe what is really missing is:
`statsd_measure_success` that would keep track of success in a tag
`statsd_measure_if` that would track only success

That leaves us with `statsd_measure` that should track everything.

Existing code point in the direction I want, so I'm using it as my argument. I'm open to changing how the `statsd_count` to ignore failures if we prefer to go the other direction. Breaking changes either way.

This gets my :+1: if we are considering "making `statsd_measure` behave the same way as `statsd_count`" good enough as a first step (even if we eventually would like to go back).

wvanbergen:
If you think this is the right way to go, I am on board.

We just need to make sure to keep the `duration` method in place. Rather than replacing it, let's add `current_timestamp` (instead of `gettime` which I don't find very Ruby-like), and rewrite duration to use it.

dnlserrano:
[This is embarassing](https://github.com/Shopify/statsd-instrument/pull/112)... I just noticed you guys had this PR open... :man_facepalming:

dnlserrano:
I'm sorry if start to sound like a broken record, but when can we expect to have this merged in? 😇 

Thanks guys! ❤️

lib/statsd/instrument.rb
```diff
@@ -277,17 +279,25 @@ def backend
...
-    metric = collect_metric(hash_argument(metric_options).merge(type: :ms, name: key, value: value))
-    result = metric unless block_given?
-    result
+    if block_given?
+      result = nil
+      start = StatsD::Instrument.current_timestamp
+      begin
+        result = yield
+      ensure
+        value = 1000 * StatsD::Instrument.current_timestamp - start
```
wvanbergen: Can you add parentheses here? `1000 * (StatsD::Instrument.current_timestamp - start). I am not too sure what the rules are in Ruby, with parentheses it's clear that it will be correct.

maximebedard:
I've updated this PR, but I've also started building the new API here: https://github.com/Shopify/statsd-instrument/pull/114 for anyone interested

dnlserrano:
Hey guys, it's me again. 😊 While we don't have the final version of the new API that @maximebedard is working on, can we get this PR merged and release a new minor of the gem?

This would help me a lot to get better metrics in the short to mid-term. I can also try to help on a solution for a more definitive way of doing things, if that's needed. Thanks! ❤️

jcmfernandes:
I was bitten by this today as well. Can we please get this PR merged and have a minor version released? :pray: 

Thank you!

PR #115, opened 2018-01-29
Support Added for collectd implementation with custom tags

shobhits147:
I have added a few lines which will enable statsd-instrument to collect metrics with custom tags from collectd as well.


wvanbergen:
Thanks for your contribution, but I will close this PR. Similar to #70, this will need a rewrite in order to be compatible with v3 of this library. I am happy to support anyone that wants to add CollectD support to the new version if anyone is willing to work on it.

Issue #116, opened 2018-03-13
Setting non-default backend for environment


btoconnor:
Hello,

I'm trying to use non-default backends from the default settings for `statsd-instrument`, and as best I can tell they're being overwritten.  It would appear that they are being overwritten here: https://github.com/Shopify/statsd-instrument/blob/master/lib/statsd/instrument/railtie.rb#L11-L13

as that is not checking for whether or not the backend is already set, it's simply setting it to be the defaults.  I've tried making an initializer that sets `StatsD.backend = StatsD::Instrument::Backends::NullBackend.new`.  After the application loads, it goes back to being set as `StatsD::Instrument::Backends::LoggerBackend`.  I've also tried setting this in `config/environments/development.rb`, but upon inspection in a console after the application loads it's still `LoggerBackend`.

I did come across issue #73 , so I tried interacting (by simply calling `StatsD.backend`) with the backend prior to changing it, which did not yield the desired results either.

What's the correct procedure for setting a new default backend for an environment?

Apologies if this is in the documentation somewhere, but I searched and could not find any references outside of simply setting `StatsD.backend`, which is not working properly for me.

btoconnor:
Any feedback on this? I'm happy to open a pull request so that the initializer doesn't overwrite settings if they exist, but I'd like to confirm that I'm not missing something obvious before embarking down that path.

mike-stewart:
I'm having this same problem.

sonya:
@btoconnor are you still having this issue? I've been setting `StatsD.backend` in an initializer and I believe it's been working fine

ekowcharles:
Same problem here. I am running this in development but will like to pipe all stats to an instance of statsd and graphite running locally. There seems not to be any way to override the backend when running in development environment.

PR #117, opened 2018-03-26
Only use the default backend for an environment if it wasn't already …

btoconnor:
…set.

This is in reference to #116 .

This pull request makes it so that the StatsD railtie initializer does not overwrite the StatsD backend if there was one already declared by the application.

Note: I'm not sure how to write a test for a railtie initializer.  I'm happy to write it if we think this is necessary and someone wants to give me a little guidance.

Pinging @jstorimer and @wvanbergen (as per contributing guidelines)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@ This file documents the changes between releases of this library. When creating 
 please at an entry to the "unreleased changes" section below.
 
 ### Unreleased changes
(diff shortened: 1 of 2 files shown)


wvanbergen:
With the major rewrite of v3, this change is no longer relevant. You can instantiate different copies of `StatsD::Instrument::Client` rather than fiddling with the default configuration for the singleton.

PR #119, opened 2018-04-10
Allow whitelisting what tags to send to datadog

singpolyma-shopify:
Whitelist matches by prefix because most tags are of the form "thing:value"


lib/statsd/instrument/backends/udp_backend.rb
```diff
@@ -76,6 +77,11 @@ def generate_packet(metric)
       command
     end
 
+    def packet_tags(tags)
+      tags = tags.select { |tag| self.tags_whitelist.any? { |white| tag.start_with?(white) } } if self.tags_whitelist
```
jasonhl: I _think_ this will let through a tag like "hostname" if the whitelist has an entry for "host" because of the `start_with?`.  Probably worth adding a test for this case.

test/udp_backend_test.rb
```diff
@@ -115,6 +115,13 @@ def test_warn_when_using_tags_and_not_on_datadog
     StatsD.increment('fooc', tags: ['ignored'])
   end
 
+  def test_support_tags_whitelist_on_datadog
```
jasonhl: Can you add a test which rejects a tag that is not in the whitelist as well?

lib/statsd/instrument/backends/udp_backend.rb
singpolyma-shopify: Whitelist should have entry for `host:` to prevent this

test/udp_backend_test.rb
singpolyma-shopify: I believe this test does that?  I have changed it to host and hostname to test the possible prefix collision there

lib/statsd/instrument/backends/udp_backend.rb
pallan: `host` is a special differentiating tag in DD if it is not submitted metrics will still work but if 2 samples from the same metric arrive with the same timestamp without a host tag, last write wins.

pallan:
How is this is going to work? We have thousands of tags going into DD today. How are they going to be added the whitelist? Do they need to be?

singpolyma-shopify:
@wvanbergen if you don't add anything to the whitelist (which existing applications will not), this change is a single extra compare against nil

singpolyma-shopify:
Benchmark:

```
Old code, no whitelist
  1.570000   0.010000   1.580000 (  1.592852)
New code, no whitelist
  1.600000   0.010000   1.610000 (  1.615217)
New code, with whitelist
  2.240000   0.010000   2.250000 (  2.244875)
```

lib/statsd/instrument/backends/udp_backend.rb
```diff
@@ -9,6 +9,7 @@ class UDPBackend < StatsD::Instrument::Backend
 
     attr_reader :host, :port
     attr_accessor :implementation
+    attr_accessor :tags_whitelist
```
wvanbergen: Can we force this to be a set, to ensure fast lookups?

lib/statsd/instrument/backends/udp_backend.rb
```diff
@@ -76,6 +77,15 @@ def generate_packet(metric)
       command
     end
 
+    def packet_tags(tags)
+      if self.tags_whitelist
+        tags = tags.select { |tag| self.tags_whitelist.any? { |white| tag.start_with?(white) } }
```
wvanbergen: I'd really prefer this to be `tags_whitelist.include?(something)` ion combination with using a set, which will be a lot faster, especially for a large whitelist.

singpolyma-shopify: If we could do that we could even use `tags &= self.tags_whitelist` -- that was my original plan, before I realised that the useful tags are all dynamic things where we only know the prefix statically

wvanbergen: you mean key/value tags, like "host:foo.bar"

wvanbergen:
What I am missing here is the rationale for this feature. This makes it a bit hard to see if the performance hit is worth it. What problems can we solve with this?

lib/statsd/instrument/backends/udp_backend.rb
singpolyma-shopify: yes

test/udp_backend_test.rb
```diff
@@ -115,6 +115,13 @@ def test_warn_when_using_tags_and_not_on_datadog
...
+    @backend.implementation = :datadog
+    @backend.tags_whitelist = ['host:']
+    @backend.expects(:write_packet).with("fooc:3|c|#host:1")
+    StatsD.increment('fooc', 3, tags: ['hostname:2', 'host:1'])
```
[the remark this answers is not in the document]
singpolyma-shopify: yes, that is what this tests :)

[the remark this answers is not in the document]
jasonhl: Oy. Sorry. I apparently can't read real well :|

PR #131, opened 2018-07-05
Allow Distribution to accept a block

pallan:
This updates the `distribution` metric type to accept a block that is very similar to how `measure` works. It also refactors the instrumentation methods to make them more DRY.


pallan:
Added the ability to define a distribution as a class method like a measure

Issue #133, opened 2018-07-17
Using return or raising exception causes metric collection to be skipped in distribution/measure


Bringer128:
Ruby interpreter version: `2.4.4p296`
The statsd-instrument version. `v2.3.0.beta5`
The StatsD backend you are using. `UDPBackend`

Using StatsD.distribution or StatsD.measure it is possible to skip the metric collection. This can happen if an exception is raised or if you return from the block.

An example of it working successfully:
```ruby
def expensive_method
  StatsD.measure('my_metric') do
    result = expensive_operation
    result.success
  end
end
```
An example of it breaking:
```ruby
def expensive_method_broken_metrics
  StatsD.measure('my_metric') do
    result = expensive_operation
    return true if result.success # Skips metric collection
    false
  end
end
```

The fix would be something like:
```ruby
def measure(key, value = nil, *metric_options, &block)
  value, metric_options = parse_options(value, metric_options)
  type = (!metric_options.empty? && metric_options.first[:as_dist] ? :d : :ms)

  result = nil
  value  = 1000 * StatsD::Instrument.duration { result = block.call } if block_given?
ensure # Add this
  metric = collect_metric(type, key, value, metric_options)
  result = metric unless block_given?
  return result # Make sure this value is returned from measure
end
```

Bringer128:
Note this might be fixed by https://github.com/Shopify/statsd-instrument/pull/111/

PR #134, opened 2018-07-31
Collect measure and distribution metrics even on return or exception

Bringer128:
Duplicate of https://github.com/Shopify/statsd-instrument/pull/111 (except no conflicts) and fixes https://github.com/Shopify/statsd-instrument/issues/133

This is a fix for when doing a return inside a `StatsD.measure` or `StatsD.distribution` call, where the metrics would not be collected.

Note that as discussed in https://github.com/Shopify/statsd-instrument/pull/111 if exceptions happen you will likely get a different distribution of performance. At the very least we should consider documenting this behaviour.

Keen to get this merged as I lost a day to this problem (existing code using a return, and I added a block to measure performance).

### DIFF
--- a/lib/statsd/instrument.rb
+++ b/lib/statsd/instrument.rb
@@ -58,17 +58,13 @@ module StatsD
 
     if Process.respond_to?(:clock_gettime)
       # @private
-      def self.duration
-        start = Process.clock_gettime(Process::CLOCK_MONOTONIC)
-        yield
-        Process.clock_gettime(Process::CLOCK_MONOTONIC) - start
+      def self.current_timestamp
+        Process.clock_gettime(Process::CLOCK_MONOTONIC)
       end
     else
       # @private
-      def self.duration
-        start = Time.now
-        yield
-        Time.now - start
+      def self.current_timestamp
+        Time.now
       end
     end
 
@@ -310,11 +306,16 @@ module StatsD
     value, metric_options = parse_options(value, metric_options)
     type = (!metric_options.empty? && metric_options.first[:as_dist] ? :d : :ms)
 
-    result = nil
-    value  = 1000 * StatsD::Instrument.duration { result = block.call } if block_given?
-    metric = collect_metric(type, key, value, metric_options)
-    result = metric unless block_given?
-    result
+    return collect_metric(type, key, value, metric_options) unless block_given?
+
+    start = StatsD::Instrument.current_timestamp
+    begin
+      block.call
+    ensure
+      # Ensure catches both a raised exception and a return in the invoked block
+      value = 1000 * (StatsD::Instrument.current_timestamp - start)
+      collect_metric(type, key, value, metric_options)
+    end
   end
 
   # Emits a counter metric.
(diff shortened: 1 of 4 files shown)


pallan:
cc @tjoyal @wvanbergen

lib/statsd/instrument.rb
```diff
@@ -373,10 +383,17 @@ def histogram(key, value, *metric_options)
   #      end
   def distribution(key, value=nil, *metric_options, &block)
     value, metric_options = parse_options(value, metric_options)
+
     result = nil
-    value  = 1000 * StatsD::Instrument.duration { result = block.call } if block_given?
-    metric = collect_metric(:d, key, value, metric_options)
-    result = metric unless block_given?
+    start = StatsD::Instrument.current_timestamp
```
tjoyal: I am not a huge fan of all these `if block_given?` even if arguably they were somewhat already present.

But this code path could only get called when a block is given (not used otherwise)

I've tried to come up with a more declarative split of the 2 flows (block vs no-block), what do you think?

```ruby
def distribution(key, value = nil, *metric_options, &block)
  value, metric_options = parse_options(value, metric_options)
  result = nil

  if block_given?
    start = StatsD::Instrument.current_timestamp
    begin
      result = block.call
    ensure
      value = 1000 * (StatsD::Instrument.current_timestamp - start)
      collect_metric(:d, key, value, metric_options)
    end
  else
    result = collect_metric(:d, key, value, metric_options)
  end

  result
end
```

lib/statsd/instrument.rb
```diff
@@ -58,20 +58,23 @@ def self.generate_metric_name(metric_name, callee, *args)
...
-        start = Time.now
-        yield
-        Time.now - start
+      def self.current_timestamp
+        Time.now
       end
     end
 
+    # @private
+    def self.duration
```
tjoyal: `duration` would then never get used anymore?

lib/statsd/instrument.rb
Bringer128: I like this!

Bringer128: I tweaked it a bit more as it's easier to return early `unless block_given?`, have another look!

lib/statsd/instrument.rb
Bringer128: Removed

Bringer128:
I'll try get some time to look into those tests tomorrow

Bringer128:
If I can get a 👍 from @tjoyal I think we can merge this?

lib/statsd/instrument/version.rb
```diff
@@ -1,5 +1,5 @@
 module StatsD
   module Instrument
-    VERSION = "2.3.2"
+    VERSION = "2.3.3"
```
tjoyal: We should probably update the changelog.

Bringer128:
> I'm not saying to not ship this, but it need to be documented so when the gem get bumped people are aware (This could trip pagers).

If this is a more significant change should we modify the minor or even major version? While it's not a breaking change at the code level (re: semver) it is potentially a breaking behaviour change that people should be aware of when they upgrade.

tjoyal:
> If this is a more significant change should we modify the minor or even major version?

I don't really have a strong opinion on this, I'm happy with a log entry and a version change.

I'm a big fan of simple, ship fast and iterate. This changeset might be a problem for some, but for others it might surface some unknown problem. I'd say let's keep the train going and simply ship & iterate.

In the end, you should always be reading at least the changelog when updating a gem. Weither it's a minor or major version number change, since we do not manage multiple version in parallel, you can only pick 2 choices: upgrade or not.

Change in the end always risk breaking someone experience, as long that we make it better for most I think we should keep going and rectify on feedback if there is any.

Bringer128:
👍  Can you review my changelog? I wanted to give a quick warning in there related to our discussions

PR #138, opened 2018-10-25
Adding synchronize to invalidate_socket

zachmccormick:
Without adding a synchronize block here, if one thread enters the
write_packet synchronize block and calls :socket, it can finish the
@socket.connect(host, port) call then pass control on to another thread,
which may be inside of the subsequent rescue block that calls
:invalidate_socket. When control goes back to the original thread, the
@socket is now nil, but will be returned to the previous frame, causing
the :send method to be called on NilClass, which will convert the
command into a symbol and try calling that as a method on the nil
object. This is avoided in this commit by wrapping the code that sets
@socket to nil in a synchronized block as well to guarantee that @socket
is never set to nil while a thread is executing a synchronize block.

The alternative here would be to wrap the entire begin/rescue block in write_packet in a synchronized block, but technically :invalidate_socket could be called elsewhere, so this made the most sense to me.

### DIFF
--- a/lib/statsd/instrument/backends/udp_backend.rb
+++ b/lib/statsd/instrument/backends/udp_backend.rb
@@ -150,7 +150,9 @@ module StatsD::Instrument::Backends
     end
 
     def invalidate_socket
-      @socket = nil
+      synchronize do
+        @socket = nil
+      end
     end
   end
 end


zachmccormick:
Relevant redacted stack trace when we experienced this today in huge numbers:
```
NoMethodError: undefined method `redacted-statsd-key:1|c' for nil:NilClass
  from statsd/instrument/backends/udp_backend.rb:140:in `block in write_packet'
  from monitor.rb:211:in `mon_synchronize'
  from statsd/instrument/backends/udp_backend.rb:139:in `write_packet'
  from statsd/instrument/backends/udp_backend.rb:111:in `collect_metric'
  from statsd/instrument.rb:456:in `collect_metric'
  from statsd/instrument.rb:332:in `increment'
```

zachmccormick:
Also I would love to see someone write a red-green test for this. I probably spent 30 minutes trying to come up with a way to do this without significant stubbing and came out empty handed

wvanbergen:
Hello @zachmccormick 

Sorry for the very late reply. This change makes sense to me, but can you rebase on the latest PR? You may also have to sign the Shopify CLA.

wvanbergen:
Actually, it turns out this has been fixed already in #147. Closing this.

PR #148, opened 2018-11-27
Allow :prefix option on metrics creation

timothysmith0609:
Closes #69 

Allows clients to pass in `:prefix` as an option to a metric. This allows classes that extend StatsD to avoid polluting the global namespace. The motivation for this is [KubernetesDeploy #384](https://github.com/Shopify/kubernetes-deploy/pull/384), though it seems it's not the first app to feel the need for this.

The change is backwards compatible and purely opt-in

### DIFF
--- a/Rakefile
+++ b/Rakefile
@@ -2,7 +2,7 @@ require 'bundler/gem_tasks'
 require 'rake/testtask'
 
 Rake::TestTask.new('test') do |t|
-  t.ruby_opts << '-rubygems'
+  t.ruby_opts << '-r rubygems'
   t.libs << 'lib' << 'test'
   t.test_files = FileList['test/*.rb']
 end
--- a/lib/statsd/instrument.rb
+++ b/lib/statsd/instrument.rb
@@ -448,11 +448,16 @@ module StatsD
 
   # Instantiates a metric, and sends it to the backend for further processing.
   # @param options (see StatsD::Instrument::Metric#initialize)
-  # @return [StatsD::Instrument::Metric] The meric that was sent to the backend.
+  # @return [StatsD::Instrument::Metric] The metric that was sent to the backend.
   def collect_metric(type, name, value, metric_options)
     value, metric_options = parse_options(value, metric_options)
-
-    options = hash_argument(metric_options).merge(type: type, name: name, value: value)
+    metric_options = hash_argument(metric_options)
+    options = metric_options.merge(
+      type: type,
+      name: name,
+      value: value,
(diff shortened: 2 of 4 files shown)


Rakefile
```diff
@@ -2,7 +2,7 @@ require 'bundler/gem_tasks'
 require 'rake/testtask'
 
 Rake::TestTask.new('test') do |t|
-  t.ruby_opts << '-rubygems'
+  t.ruby_opts << '-r rubygems'
```
karanthukral: unrelated change>

lib/statsd/instrument/metric.rb
```diff
@@ -47,7 +49,11 @@ def initialize(options = {})
     @type = options[:type] or raise ArgumentError, "Metric :type is required."
     @name = options[:name] or raise ArgumentError, "Metric :name is required."
     @name = normalize_name(@name)
-    @name = StatsD.prefix ? "#{StatsD.prefix}.#{@name}" : @name unless options[:no_prefix]
+    @name = if options[:prefix]
```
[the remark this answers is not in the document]
timothysmith0609: `prefix` will override (which I should document).
I could be persuaded to have `:no_prefix` have greater precedence but it's 6 of 1, half-dozen of the other.

Rakefile
timothysmith0609: CI was complaining about this (and was in fact preventing me from testing locally)

lib/statsd/instrument.rb
```diff
@@ -448,11 +448,16 @@ def parse_options(value, metric_options)
...
   def collect_metric(type, name, value, metric_options)
     value, metric_options = parse_options(value, metric_options)
-
-    options = hash_argument(metric_options).merge(type: type, name: name, value: value)
+    metric_options = hash_argument(metric_options)
+    options = metric_options.merge(
+      type: type,
+      name: name,
+      value: value,
+      prefix: metric_options[:prefix]
```
fw42: what does this line actually do? I'm reading this as `metric_options.merge(prefix: metric_options[:prefix])`, which doesn't look like it's doing anything, is it?

lib/statsd/instrument/metric.rb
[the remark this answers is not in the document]
fw42: Hm. `no_prefix` having precedence sounds less surprising to me.

lib/statsd/instrument.rb
timothysmith0609: Ah, you are right I goofed

PR #151, opened 2019-03-25
Add the latest supported ruby version to RVM version matrix

effron:
All of the currently supported ruby versions are officially [EOL](https://www.ruby-lang.org/en/downloads/branches/) (or almost EOL) , so this adds CI testing for the latest stable ruby version. Should we also add testing for 2.4.x and 2.5.x?

I can also add something to the changelog if this requires any actual code changes!


agirlnamedsophia:
maybe update the "compatibility" section too? https://github.com/Shopify/statsd-instrument#compatibility

effron:
@agirlnamedsophia updated!

PR #152, opened 2019-08-15
Remove unnecessary array allocation from use of map

thegedge:
While profiling one of our libraries with stackprof, we noticed a large number of samples ended up in `Metric#to_s`. It was obvious that this is largely due to the array allocation that occurs with `map`. To avoid this allocation, we simply use `each` instead.

My own benchmarks for this function:

```ruby
                to_s    54.756k i/100ms
               to_s2    76.205k i/100ms
                to_s    652.458k (± 1.8%) i/s -      3.285M in   5.037082s
               to_s2    931.104k (± 2.6%) i/s -      4.725M in   5.077774s

Comparison:
               to_s2:   931104.4 i/s
                to_s:   652458.3 i/s - 1.43x  slower
```

where `to_s` is the old implementation and `to_s2` is the new implementation in this PR, the benchmark tested against a metric with two tags.

### Notes

Weirdly enough, the above benchmark looks like we're only shaving off less than a microsecond, which is not really saving much unless this is getting called A LOT in a request lifecycle. I observed some slowness once in my own profiling:

Screen Shot 2019-08-15 at 4 47 58 PM

but I haven't been able to reproduce this slowness though, so it may have been a one-off, and perhaps we can simply ignore this PR.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,8 @@ please at an entry to the "unreleased changes" section below.
 
 ### Unreleased changes
 
+- Improve performance of `Metric#to_s`
+
 ## Version 2.3.3
 
 - Capture measure and distribution metrics on exception and early return (#134)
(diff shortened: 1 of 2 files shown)


thegedge:
I just realized this is for the logger backend, so wouldn't really have an impact on production (I doubt anyone uses the logging backend in production...).

PR #153, opened 2019-08-26
Fix bug in escaping newlines for events with Datadog Backend

coreyja:
Ran into this issue where we were sending events with newlines that were
NOT getting properly escaped.

This was due to a few compounding issues:

1. `'\n'` is NOT the same as `"\n"` in Ruby. Single quotes do NOT treat
   `\` special so that first string is 2 chars. While the double quoted
   string takes `\n` to mean the newline character
1. `tr` only ever replaces single characters. The usage of `tr` does not
   appear to be correct. Since the intended behavior is to take a single
   char `\n` (the new line character) and replace it with TWO chars, `\`
   followed by an `n`, or `"\\n"` in Ruby. Since `tr` can only replace
   single chars it does not seem well suited for this problem

This replaces the usage of `tr` with `gsub` which has the right
behavior.

It also updates the tests to use double quotes, since their use of
single quotes appears to be hiding this bug.

Let me know if there is anything else I can do to help get this merged into this Upstream repo! Thanks!

### DIFF
--- a/lib/statsd/instrument/backends/udp_backend.rb
+++ b/lib/statsd/instrument/backends/udp_backend.rb
@@ -27,8 +27,8 @@ module StatsD::Instrument::Backends
         packet = ""
 
         if metric.type == :_e
-          escaped_title = metric.name.tr('\n', '\\n')
-          escaped_text = metric.value.tr('\n', '\\n')
+          escaped_title = metric.name.gsub("\n", "\\n")
+          escaped_text = metric.value.gsub("\n", "\\n")
 
           packet << "_e{#{escaped_title.size},#{escaped_text.size}}:#{escaped_title}|#{escaped_text}"
           packet << generate_metadata(metric, EVENT_OPTIONS)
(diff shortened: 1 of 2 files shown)


coreyja:
@wvanbergen I have signed the CLA now, but am unsure what the best way to kick the bot again is!
Is that something you could assist with?

wvanbergen:
It's supposed to update. Let me investigate...

coreyja:
Thanks! I know I submitted it cause I went to redo it this morning and got told I was already signed 🙃

wvanbergen:
Haven't heard back from the people that maintain the CLA tool.
In the mean time, one thing you could try is to rebase this PR so it will re-evaluate the CLA status.

coreyja:
@wvanbergen Rebase seemed to work! It just needed a little kick lol

wvanbergen:
Great success 🎉 

Thanks for your contribution!

coreyja:
Thanks again for the review @wvanbergen and @maximebedard!

coreyja:
@wvanbergen Is there any chance we could get a release cut with these changes?

I can also get a version bump PR out if that would be helpful in that process!

wvanbergen:
Released as 2.3.4

Issue #155, opened 2019-09-04
Add option to intercept tags


fabiormoura:
It would be nice to be able to attach or modify tags via an interceptor. The most common use case this is useful is when common tags(s) are used everywhere in code.

For instance, recently our platform team introduced the concept of `runtime` and they provided an interface to identify which `runtime` our service is currently running onto as in `Runtime.current`. 

We decided to add the runtime tag to all our metrics but we did that by monkey-patching [`StatsD::Metric#initialize`](https://github.com/Shopify/statsd-instrument/blob/808f62362c5acaf2caaa99d88ae4b8f39569e8aa/lib/statsd/instrument/metric.rb#L49) to insert the `runtime` tag into `@tags` as the snippet below shows:

```ruby

    RUNTIME_ID_PREFIX = "runtime_id:"

    def initialize(*)
      super
            @tags ||= []

      unless @tags.any? { |tag| tag.start_with?(RUNTIME_ID_PREFIX) }
        @tags << "#{RUNTIME_ID_PREFIX}#{Runtime.current}"
      end
    end

```

What are your thoughts to allow passing a new proc to StatsD and then use it for intercepting `tags` before `metric` is instantiated at https://github.com/Shopify/statsd-instrument/blob/808f62362c5acaf2caaa99d88ae4b8f39569e8aa/lib/statsd/instrument.rb#L462

```ruby
StatsD.tags_appender = proc do |tags|
  tags[:runtime_id] = Runtime.current
end
```

wvanbergen:
The main hesitations I have that this would be a massive slowdown. StatsD calls are part of some extremely hot code paths, and introducing a `proc` for every call sounds very expensive

PR #159, opened 2019-09-16
Add default_tags attribute to StatsD module

fabiormoura:
Implements: #155 

Adds new attribute to `StatsD` module allowing to specify default tags that will be applied to all metrics.

### DIFF
--- a/lib/statsd/instrument.rb
+++ b/lib/statsd/instrument.rb
@@ -284,8 +284,13 @@ module StatsD
     end
   end
 
-  attr_accessor :logger, :default_sample_rate, :prefix, :default_tags
+  attr_accessor :logger, :default_sample_rate, :prefix
   attr_writer :backend
+  attr_reader :default_tags
+
+  def default_tags=(tags)
+    @default_tags = StatsD::Instrument::Metric.normalize_tags(tags)
+  end
 
   def backend
     @backend ||= StatsD::Instrument::Environment.default_backend
--- a/lib/statsd/instrument.rb
+++ b/lib/statsd/instrument.rb
@@ -28,6 +28,10 @@ require 'logger'
 #   The logger to use in case of any errors. The logger is also used as default logger
 #   for the LoggerBackend (although this can be overwritten).
 #
+# @!attribute default_tags
+#   The tags to apply to all metrics.
+#   @return [Array<String>, Hash<String, String>, nil] The default tags, or <tt>nil</tt> when no default tags is used
+#
 #   @see StatsD::Instrument::Backends::LoggerBackend
 #   @return [Logger]
 #
@@ -280,7 +284,7 @@ module StatsD
     end
   end
 
-  attr_accessor :logger, :default_sample_rate, :prefix
+  attr_accessor :logger, :default_sample_rate, :prefix, :default_tags
   attr_writer :backend
 
   def backend
--- a/lib/statsd/instrument/metric.rb
+++ b/lib/statsd/instrument/metric.rb
@@ -72,7 +72,7 @@ class StatsD::Instrument::Metric
     @sample_rate = options[:sample_rate] || StatsD.default_sample_rate
     @tags = StatsD::Instrument::Metric.normalize_tags(options[:tags])
     if StatsD.default_tags
-      @tags = Array(@tags) | StatsD.default_tags
+      @tags = Array(@tags) + StatsD.default_tags
     end
     @metadata = options.reject { |k, _| [:type, :name, :value, :sample_rate, :tags].include?(k) }
   end
--- a/lib/statsd/instrument/metric.rb
+++ b/lib/statsd/instrument/metric.rb
@@ -62,7 +62,7 @@ class StatsD::Instrument::Metric
     @sample_rate = options[:sample_rate] || StatsD.default_sample_rate
     @tags        = StatsD::Instrument::Metric.normalize_tags(options[:tags])
     if StatsD.default_tags
-      @tags      = Array(@tags) | StatsD::Instrument::Metric.normalize_tags(StatsD.default_tags)
+      @tags      = Array(@tags) | StatsD.default_tags
     end
     @metadata    = options.reject { |k, _| [:type, :name, :value, :sample_rate, :tags].include?(k) }
   end
--- a/lib/statsd/instrument/metric.rb
+++ b/lib/statsd/instrument/metric.rb
@@ -61,6 +61,9 @@ class StatsD::Instrument::Metric
     @value       = options[:value] || default_value
     @sample_rate = options[:sample_rate] || StatsD.default_sample_rate
     @tags        = StatsD::Instrument::Metric.normalize_tags(options[:tags])
+    if StatsD.default_tags
+      @tags      = Array(@tags) | StatsD::Instrument::Metric.normalize_tags(StatsD.default_tags)
+    end
     @metadata    = options.reject { |k, _| [:type, :name, :value, :sample_rate, :tags].include?(k) }
   end
 
--- a/test/benchmark/default_tags.rb
+++ b/test/benchmark/default_tags.rb
@@ -1,3 +1,5 @@
+# frozen_string_literal: true
+
 require 'statsd-instrument'
 require 'benchmark/ips'
 
@@ -6,19 +8,15 @@ StatsD.logger = Logger.new('/dev/null')
 class Suite
   def warming(*args)
     StatsD.default_tags = if args[0] == "with default tags"
-                            {:first_tag => 'first_value', :second_tag => 'second_value'}
-                          else
-                            nil
-                          end
+      { first_tag: 'first_value', second_tag: 'second_value' }
+    end
     puts "warming with default tags: #{StatsD.default_tags}"
   end
 
   def running(*args)
     StatsD.default_tags = if args[0] == "with default tags"
-                            {:first_tag => 'first_value', :second_tag => 'second_value'}
-                          else
-                            nil
-                          end
+      { first_tag: 'first_value', second_tag: 'second_value' }
+    end
     puts "running with default tags: #{StatsD.default_tags}"
   end
 
@@ -32,13 +30,17 @@ end
 suite = Suite.new
 
 Benchmark.ips do |bench|
-  bench.config(:suite => suite)
+  bench.config(suite: suite)
   bench.report("without default tags") do
-    StatsD.increment('GoogleBase.insert', tags: { :first_tag => 'first_value', :second_tag => 'second_value', :third_tag => 'third_value' })
+    StatsD.increment('GoogleBase.insert', tags: {
+      first_tag: 'first_value',
+      second_tag: 'second_value',
+      third_tag: 'third_value',
+    })
   end
 
   bench.report("with default tags") do
-    StatsD.increment('GoogleBase.insert', tags: { :third_tag => 'third_value' })
+    StatsD.increment('GoogleBase.insert', tags: { third_tag: 'third_value' })
   end
 
   bench.compare!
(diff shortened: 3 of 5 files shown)


lib/statsd/instrument/metric.rb
```diff
@@ -61,6 +61,9 @@ def initialize(options = {})
     @value       = options[:value] || default_value
     @sample_rate = options[:sample_rate] || StatsD.default_sample_rate
     @tags        = StatsD::Instrument::Metric.normalize_tags(options[:tags])
+    if StatsD.default_tags
+      @tags        = Array(@tags) | StatsD::Instrument::Metric.normalize_tags(StatsD.default_tags)
```
wvanbergen: For performance reasons, can we normalize `default_tags` when you assign them, rather than for every metric that gets emitted?

fabiormoura:
> Can you write a simple benchmark, to compare metric emission performance:
> 
> 1. Before this change
> 2. After this change, with no default tags
> 3. After this change, with default tags set.

I'll write the benchmark once the implementation satisfies. I'm making the change you suggested on the other comment.

lib/statsd/instrument/metric.rb
wvanbergen: One question is how we want to handle duplicate tags. How do we want to handle a situation when we have a tag that conflicts with a default tag (e.g. `foo:bar` and `foo:baz`). My instinct tells me that the value in `tags` should overwrite the value in `default_tags`.

We cannot really check for duplicates here for performance reasons. However, we can influence the ordering. Can you check what value will be used if you have two tags with the same key: the first or the last one? Based on that, we can determine whether we want to do `tags | default_tags` or `default_tags | tags`.

wvanbergen:
> I'll write the benchmark once the implementation satisfies.

Performance is a feature. The implementation only satisfies if the performance is good enough.  :)

lib/statsd/instrument/metric.rb
fabiormoura: The way this is currently implemented does not handle tags with duplicated keys and the resulting list will include both unless their "key:value" are the same.

```ruby
options[:tags] = {"foo" => "bar"}
StatsD.default_tags = {"foo" => "baz"}

@tags = StatsD::Instrument::Metric.normalize_tags(options[:tags])
=> ["foo:bar"]
if StatsD.default_tags
  @tags        = Array(@tags) | StatsD::Instrument::Metric.normalize_tags(StatsD.default_tags)
end
=> ["foo:bar", "foo:baz"]
```

wvanbergen: So my question is: how does Datadog interpret this?

fabiormoura: It is interpreted as separate tags and ordering is irrelevant, I think. I tested that both can be used as any other tags, for instance, to narrow down searches.

```ruby
StatsD.increment("fmaia_test", tags: ["foo:bar", "foo:baz"])
```

image

image

wvanbergen:
Final question, for `StatsD.prefix`, we have a `no_prefix` option to prevent the prefix from being added to the metric name. Do we need the same for the `default_tags`?

lib/statsd/instrument/metric.rb
wvanbergen: I guess this is acceptable

fabiormoura: Sure. I'll add this to the PR.

fabiormoura:
> Final question, for `StatsD.prefix`, we have a `no_prefix` option to prevent the prefix from being added to the metric name. Do we need the same for the `default_tags`?

It is up to you. To me not having this for now is fine as well. If someone else needs that as a requirement they could contribute through PR. But if you believe that should be part of this PR, let me know.

fabiormoura:
> Can you write a simple benchmark, to compare metric emission performance:
> 
> 1. Before this change
> 2. After this change, with no default tags
> 3. After this change, with default tags set.

How to run it:

```
bundle exec ruby test/benchmark/default_tags.rb
```

The results are:

> 1. Before this change

```
without default tags     7.050k i/100ms
without default tags     73.708k (± 2.3%) i/s -    373.650k in   5.072079s
```

> 2. After this change, with no default tags
> 3. After this change, with default tags set.

```
without default tags     7.142k i/100ms
   with default tags     4.560k i/100ms
without default tags     49.446k (± 5.5%) i/s -    249.970k in   5.073286s
   with default tags     48.173k (± 1.5%) i/s -    241.680k in   5.018064s

Comparison:
without default tags:    49445.8 i/s
   with default tags:    48172.8 i/s - same-ish: difference falls within error
```

lib/statsd/instrument/metric.rb
scalp42: @fabiormoura is that DataDog?

fabiormoura: > @fabiormoura is that DataDog?

It is @scalp42

fabiormoura:
> > Can you write a simple benchmark, to compare metric emission performance:
> > 
> > 1. Before this change
> > 2. After this change, with no default tags
> > 3. After this change, with default tags set.
> 
> How to run it:
> 
> ```
> bundle exec ruby test/benchmark/default_tags.rb
> ```
> 
> The results are:
> 
> > 1. Before this change
> 
> ```
> Warming up --------------------------------------
> without default tags     5.040k i/100ms
> Calculating -------------------------------------
> without default tags     52.150k (± 1.2%) i/s -    262.080k in   5.026255s
> ```
> 
> > 1. After this change, with no default tags
> > 2. After this change, with default tags set.
> 
> ```
> Warming up --------------------------------------
> warming with default tags:
> without default tags     4.963k i/100ms
> warming with default tags: {:first_tag=>"first_value", :second_tag=>"second_value"}
>    with default tags     4.838k i/100ms
> Calculating -------------------------------------
> running with default tags:
> without default tags     52.946k (± 1.4%) i/s -    268.002k in   5.062786s
> running with default tags: {:first_tag=>"first_value", :second_tag=>"second_value"}
>    with default tags     50.408k (± 1.4%) i/s -    256.414k in   5.087807s
> 
> Comparison:
> without default tags:    52945.9 i/s
>    with default tags:    50408.1 i/s - 1.05x  slower
> ```

What I posted originally was incorrect as the benchmark code wasn't setup properly. I edited my original answer to reflect with correct results.

wvanbergen:
> It is up to you. To me not having this for now is fine as well. If someone else needs that as a requirement they could contribute through another PR.

That works for me :+1:

wvanbergen:
We are using set union (the `|` operator) to combine the tags. My guess this involves checking the resulting array for duplicates, which is quite computationally heavy.

Given that we don't actually filter out duplicates for key/value-based tags anyway, I think we can completely forget about removing duplicates and combine the two arrays using array concatenation (`+`) instead. Can you run the benchmark again with that change?

fabiormoura:
> We are using set union (the `|` operator) to combine the tags. My guess this involves checking the resulting array for duplicates, which is quite computationally heavy.
> 
> Given that we don't actually filter out duplicates for key/value-based tags anyway, I think we can completely forget about removing duplicates and combine the two arrays using array concatenation (`+`) instead. Can you run the benchmark again with that change?

Sorry for not giving you the entire context. The code I benchmarked did not have the change you proposed here: https://github.com/Shopify/statsd-instrument/pull/159#discussion_r324825072.

Let me first make this change and then benchmark it again before deciding on what operator to use `|` vs `+`.

wvanbergen:
Now you have the benchmark script, it'll be straightforward to test the perf impact of some of these small changes.

fabiormoura:
Below the benchmark after https://github.com/Shopify/statsd-instrument/pull/159#discussion_r324825072 has been implemented: 

```
warming with default tags:
without default tags     4.255k i/100ms
warming with default tags: ["first_tag:first_value", "second_tag:second_value"]
   with default tags     4.877k i/100ms
running with default tags:
without default tags     49.541k (± 5.0%) i/s -    251.045k in   5.081559s
running with default tags: ["first_tag:first_value", "second_tag:second_value"]
   with default tags     53.646k (± 2.8%) i/s -    268.235k in   5.004186s

Comparison:
   with default tags:    53646.1 i/s
without default tags:    49541.4 i/s - 1.08x  slower
```

fabiormoura:
The other benchmark taken after changing the operator to `+` to test if it caused any improvements:

```ruby
warming with default tags:
without default tags     4.887k i/100ms
warming with default tags: ["first_tag:first_value", "second_tag:second_value"]
   with default tags     5.341k i/100ms
running with default tags:
without default tags     52.123k (± 2.8%) i/s -    263.898k in   5.067190s
running with default tags: ["first_tag:first_value", "second_tag:second_value"]
   with default tags     56.378k (± 3.8%) i/s -    283.073k in   5.029278s

Comparison:
   with default tags:    56377.9 i/s
without default tags:    52122.6 i/s - 1.08x  slower
```

lib/statsd/instrument.rb
```diff
@@ -280,9 +284,13 @@ def method_visibility(method)
     end
   end
 
-  attr_accessor :logger, :default_sample_rate, :prefix
+  attr_accessor :logger, :default_sample_rate, :prefix, :default_tags
```
wvanbergen: Nitpick: Because you override the writer, this could be just an `attr_reader`

lib/statsd/instrument.rb
```diff
@@ -280,9 +284,13 @@ def method_visibility(method)
...
   attr_writer :backend
 
+  def default_tags=(tags)
+    @default_tags = tags ? StatsD::Instrument::Metric.normalize_tags(tags) : nil
```
wvanbergen: No need for this ternary operator, right? `normalize_tags` knows how to deal with `nil` already?

lib/statsd/instrument/metric.rb
```diff
@@ -61,6 +61,9 @@ def initialize(options = {})
     @value       = options[:value] || default_value
     @sample_rate = options[:sample_rate] || StatsD.default_sample_rate
     @tags        = StatsD::Instrument::Metric.normalize_tags(options[:tags])
+    if StatsD.default_tags
+      @tags      = Array(@tags) | StatsD.default_tags
```
wvanbergen: `StatsD.default_tags` can be `nil`. `[:a] | nil` raises a TypeError. 

You probably have to wrap `default_tags` in `Array()`, or even better, do this in the setter above.

fabiormoura: The if statement before `@tags` is re-assigned should prevent this kind of exception to happen, right?

wvanbergen: ah yes, never mind me.

wvanbergen:
It's very weird to me that in the last two benchmarks, it's faster to use `default_tags` compared to not using them. That doesn't add up for me :/

fabiormoura:
> It's very weird to me that in the last two benchmarks, it's faster to use `default_tags` compared to not using them. That doesn't add up for me :/

That is justifiable. Let's walk through that step by step:

### No default tags

The code below emits a metrics containing a hash with 3 tags. For every iteration of the test the hash has is normalized.

```
StatsD.increment('GoogleBase.insert', tags: { :first_tag => 'first_value', :second_tag => 'second_value', :third_tag => 'third_value' })
```

### With default tags
```
StatsD.default_tags = {:first_tag => 'first_value', :second_tag => 'second_value'} # set up only once before the benchmark starts
```
and then 

```
StatsD.increment('GoogleBase.insert', tags: { :third_tag => 'third_value' })
```

The difference now is that the `Hash` passed to the metric has a size of 1 and it gets normalized faster compared to the `Hash` of the scenario without default tags where its size is 3.

fabiormoura:
@wvanbergen please let me know if you think there is more work to be done here.

PR #161, opened 2019-09-20
Make statsd-instrument frozen-string-literal compatible.

wvanbergen:
This adds `# frozen_string_literal: true` to the top of every Ruby file, and addresses some incompatibilities that came to light by doing so.

### DIFF
--- a/lib/statsd/instrument/backends/udp_backend.rb
+++ b/lib/statsd/instrument/backends/udp_backend.rb
@@ -1,3 +1,5 @@
+# frozen_string_literal: true
+
 require 'monitor'
 
 module StatsD::Instrument::Backends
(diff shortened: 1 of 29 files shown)


lib/statsd/instrument/backends/udp_backend.rb
```diff
@@ -24,7 +26,7 @@ class DogStatsDProtocol
       SUPPORTED_METRIC_TYPES = BASE_SUPPORTED_METRIC_TYPES.merge(h: true, _e: true, _sc: true, d: true)
 
       def generate_packet(metric)
-        packet = ""
+        packet = StringIO.new
```
casperisfine: `StringIO` is a bit much, just do `packet = +""`

lib/statsd/instrument/metric.rb
```diff
@@ -81,10 +83,11 @@ def default_value
   # @private
   # @return [String]
   def to_s
-    str = "#{TYPES[type]} #{name}:#{value}"
+    str = StringIO.new
```
casperisfine: Same here. `+""`

lib/statsd/instrument/backends/udp_backend.rb
wvanbergen: TIL!

PR #162, opened 2019-09-20
Address Ruby warnings that come up when running `rake test` on this library

wvanbergen:
Nothing earth shattering here, but your terminal buffer will be happier for it.


test/statsd_test.rb
```diff
@@ -65,7 +65,7 @@ def test_statsd_measure_returns_return_value_of_block
     assert_equal 'sarah', return_value
   end
 
-  def test_statsd_measure_returns_return_value_of_block_even_if_nil
```
wvanbergen: This method is defined elsewhere already, with a slightly different implementation.

PR #166, opened 2019-09-20
Handle exceptions +when capturing statsd packets properly

wvanbergen:
Fixes #103.

In #103, @fw42 brought to light that the following test passes, while it really shouldn't:

``` ruby
test "foo" do
      raise "hello"
    end
  end
end
```

This is because the `assert_statsd_increment` gets short-circuited when an exception happens, and we never assert the metrics.

The first commit adds a test that shows that this behavior is indeed broken. The second commit fixes it, by doing the metric assertions in an `ensure` block.

### DIFF
--- a/test/assertions_test.rb
+++ b/test/assertions_test.rb
@@ -315,6 +315,42 @@ class AssertionsTest < Minitest::Test
     end
   end
 
+  def test_assertion_with_exceptions
+    assert_no_assertion_triggered do
+      @test_case.assert_raises(RuntimeError) do
+        @test_case.assert_statsd_increment('counter') do
+          StatsD.increment('counter')
+          raise "foo"
+        end
+      end
+    end
+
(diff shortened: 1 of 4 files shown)


test/assertions_test.rb
```diff
@@ -315,6 +315,42 @@ def test_nested_assertions
...
+      end
+    end
+
+    assert_assertion_triggered do
+      @test_case.assert_raises(RuntimeError) do
+        @test_case.assert_statsd_increment('counter') do
+          raise "foo"
+        end
+      end
+    end
```
wvanbergen: Without the second commit that fixes the issue, that assertion will fail:

```
  1) Failure:
AssertionsTest#test_assertion_with_exceptions [/Users/willem/src/github.com/Shopify/statsd-instrument/test/assertions_test.rb:345]:
No assertion was triggered, but one was expected.
```

PR #168, opened 2019-09-21
Clean up & optimize duration calculations

wvanbergen:
`Process.clock_gettime(Process::CLOCK_MONOTONIC)` is now supported on all platforms and ruby versions we care about. We don’t need a fallback anymore as a result.

### Performance

I've also looked into using `Process.clock_gettime(Process::CLOCK_MONOTONIC, :float_millisecond)` to obtain timestamps in the correct unit directly, so we don't have to do the conversion from seconds ourselves. I've added a benchmark script that gives us several ways to get a timestamp in milliseconds, which is what we need for our duration measurements:

```
Process.clock_gettime in milliseconds (int)
                          4.174M (± 1.8%) i/s -     20.966M in   5.024129s
Process.clock_gettime in milliseconds (float)
                          4.393M (± 2.9%) i/s -     22.114M in   5.038199s
Process.clock_gettime in seconds (float), multiplied by 1000
                          4.526M (± 2.1%) i/s -     22.780M in   5.035625s
Process.clock_gettime in seconds (float), multiplied by 1000.0
                          4.867M (± 2.1%) i/s -     24.461M in   5.028644s
Time.now, multiplied by 1000
                          1.791M (± 1.8%) i/s -      9.045M in   5.051360s

Comparison:
Process.clock_gettime in seconds (float), multiplied by 1000.0:  4866582.5 i/s
Process.clock_gettime in seconds (float), multiplied by 1000:  4525763.2 i/s - 1.08x  slower
Process.clock_gettime in milliseconds (float):  4393428.5 i/s - 1.11x  slower
Process.clock_gettime in milliseconds (int):  4174452.2 i/s - 1.17x  slower
Time.now, multiplied by 1000:  1791272.0 i/s - 2.72x  slower
```

Observations:
- `Time.now` is very slow compared to `clock_gettime`.
- Interestingly, converting ourselves from seconds is slightly faster than asking for the timestamps in milliseconds directly. I was not expecting this
- We can even get a small performance boost to make sure our conversion is Float x Float, rather than Int x Float, by converting our `1000` constant to `1000.0`

I've updated out conversions to use a float constant (`1000.0`).

### DIFF
--- a/lib/statsd/instrument.rb
+++ b/lib/statsd/instrument.rb
@@ -357,7 +357,7 @@ module StatsD
       block.call
     ensure
       # Ensure catches both a raised exception and a return in the invoked block
-      value = 1000 * (Process.clock_gettime(Process::CLOCK_MONOTONIC) - start)
+      value = 1000.0 * (Process.clock_gettime(Process::CLOCK_MONOTONIC) - start)
       collect_metric(type, key, value, metric_options)
     end
   end
@@ -425,7 +425,7 @@ module StatsD
     begin
       block.call
     ensure
-      value = 1000 * (Process.clock_gettime(Process::CLOCK_MONOTONIC) - start)
+      value = 1000.0 * (Process.clock_gettime(Process::CLOCK_MONOTONIC) - start)
       collect_metric(:d, key, value, metric_options)
     end
   end
--- a/lib/statsd/instrument.rb
+++ b/lib/statsd/instrument.rb
@@ -62,26 +62,26 @@ module StatsD
       metric_name.respond_to?(:call) ? metric_name.call(callee, args).gsub('::', '.') : metric_name.gsub('::', '.')
     end
 
-    if Process.respond_to?(:clock_gettime)
-      # @private
-      def self.current_timestamp
-        Process.clock_gettime(Process::CLOCK_MONOTONIC)
-      end
-    else
-      # @private
-      def self.current_timestamp
-        Time.now
-      end
+    # Even though this method is considered private, and is no longer used internally,
+    # applications in the wild rely on it. As a result, we cannot remove this method
+    # until the next major version.
+    #
+    # @deprecated Use Process.clock_gettime(Process::CLOCK_MONOTONIC) instead.
+    def self.current_timestamp
+      Process.clock_gettime(Process::CLOCK_MONOTONIC)
     end
 
     # Even though this method is considered private, and is no longer used internally,
     # applications in the wild rely on it. As a result, we cannot remove this method
     # until the next major version.
-    # @private
+    #
+    # @deprecated You can implement similar functionality yourself using
+    #   `Process.clock_gettime(Process::CLOCK_MONOTONIC)`. Think about what will
+    #   happen if an exception happens during the block execution though.
     def self.duration
-      start = current_timestamp
+      start = Process.clock_gettime(Process::CLOCK_MONOTONIC)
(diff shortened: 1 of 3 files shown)


lib/statsd/instrument.rb
```diff
@@ -62,26 +62,26 @@ def self.generate_metric_name(metric_name, callee, *args)
...
-      # @private
-      def self.current_timestamp
-        Time.now
-      end
+    # Even though this method is considered private, and is no longer used internally,
+    # applications in the wild rely on it. As a result, we cannot remove this method
+    # until the next major version.
+    #
+    # @deprecated Use Process.clock_gettime(Process::CLOCK_MONOTONIC) instead.
+    def self.current_timestamp
```
wvanbergen: @Bringer128 do you think we need to keep this method around because people depend on it in their apps? Or can we :fire: it?

Bringer128: I removed this previously and we needed to revert as people were relying on it; as the comment says we should bump the major version number if we want to remove this.

test/statsd_test.rb
```diff
@@ -52,7 +52,7 @@ def test_statsd_measure_with_explicit_value_and_sample_rate
   end
 
   def test_statsd_measure_with_benchmarked_block_duration
-    StatsD::Instrument.stubs(:current_timestamp).returns(5.0, 5.0 + 1.12)
+    Process.stubs(:clock_gettime).returns(5.0, 5.0 + 1.12)
```
Bringer128: ❤️ thanks for fixing this up so it mocks true dependencies and not internal wiring of the code.

dylanahsmith: Are we doing this in applications?  Since it can result in unintended consequences by mocking something more generic.

Bringer128: After discussing with @dylanahsmith the concern was around Shopify applications and unrelated to this public gem. There isn't an issue with this stubbing as it is in a _unit test_ for a piece of logic where the appropriate seam to stub is `Process.clock_gettime`

test/benchmark/clock_gettime.rb
```diff
@@ -0,0 +1,27 @@
...
+  bench.report("Process.clock_gettime in milliseconds (float)") do
+    Process.clock_gettime(Process::CLOCK_MONOTONIC, :float_millisecond)
+  end
+
+  bench.report("Process.clock_gettime in seconds (float), multiplied by 1000") do
+    1000 * Process.clock_gettime(Process::CLOCK_MONOTONIC)
+  end
+
+  bench.report("Process.clock_gettime in seconds (float), multiplied by 1000.0") do
+    1000.0 * Process.clock_gettime(Process::CLOCK_MONOTONIC)
```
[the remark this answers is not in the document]
wvanbergen: The underlying resolution is a struct that holds seconds & nanoseconds as separate members, stored as `time_t` & `long int`: https://linux.die.net/man/3/clock_gettime. Because of this underlying resolution, I am not too worried about additional* floating point error. (* = We have to convert this struct into a floating point number anyway, so it's already a lossy operation)

I think this also explains why `float_seconds` is faster than `float_microseconds`, because it's easier to convert this struct to a Float where seconds are the integer part. #dunnolol though

This should be the same for any CPU type. Because this is part of the POSIX standard, that characteristic should be true for all UNIX-like OSes. For Windows it may not be the case but I am happy to ignore that.

PR #170, opened 2019-09-22
Run benchmark as part of CI.

wvanbergen:
This runs the benchmark script it just added as part of CI on Github Actions.

- Run it on the current sha
- Check out master
- Run it again
- Print comparison.

As a result, we will always have some performance information that we can take into account when evaluating a pull request.

### DIFF
--- a/.github/workflows/benchmark.yml
+++ b/.github/workflows/benchmark.yml
@@ -1,0 +1,32 @@
+name: Benchmarks
+
+on: push
+
+jobs:
+  test:
+    name: Send metric over UDP
+    runs-on: ubuntu-18.04
+
+    steps:
+    - uses: actions/checkout@v1
+
+    - name: Setup Ruby
+      uses: actions/setup-ruby@v1
+      with:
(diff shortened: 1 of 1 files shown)


.github/workflows/benchmark.yml
```diff
@@ -0,0 +1,32 @@
...
+    - name: Setup Ruby
+      uses: actions/setup-ruby@v1
+      with:
+        ruby-version: 2.6
+
+    - name: Install dependencies
+      run: gem install bundler && bundle install --jobs 4 --retry 3
+
+    - name: Run benchmark on branch
+      run: benchmark/send-metrics-to-local-udp-receiver
```
dylanahsmith: Shouldn't we get rid of the test/benchmark/ folder?  It is kind of confusing to have benchmarks in the test folder, but then not being the ones run on CI

wvanbergen: Yeah, I was going to clean that up in a followup.

PR #172, opened 2019-09-23
Add a new StatsD::Instrument::Client

wvanbergen:
This sketches how I was thinking of implementing the new StatsD client.

- It is not a singleton, so you can instantiate multiple of them.
- Rather than allocating `Metric` objects that get turned into a string before being sent over the UDP socket, it immediately generates strings. 
- You can inject a DatagramBuilder class into the client, which can cover the differences between implementations, and/or being more lenient or strict when it comes to supported types, metric names, tags, etc.
- These datagram strings are sent directly to the sink (e.g. UDP socket) as is.
- The UDP sink implementation is much simpler than before. The `NullSink` and `CaptureSink` are pretty similar to the old backends of the same time.

### Migration path

For now, this new client is implemented in a completely independent way. The Ruby files are not even `require`d yet. So we can safely land this client implementation on master and iterate on it until we are ready.

Compatibility questions?
- Right now, `StatsD.instrument` returns a `StatsD::Instrument::Metric` instance. The new implementation does not use this class anymore. What should we do?
  1. Return a `StatsD::Instrument::Datagram` instance that is duck-type compatible with `StatsD::Instrument::Metric.
  2. Break backwards compatibility and the datagram as string.
  3. Assume that basically nobody really uses the return value. Break backwards compatibility and return `void`.

- Different backends/implementations support a different subset/superset of the StatsD protocol. How should we handle when the user asks for something that the implementation does not support? (E.g. tags, distribution metrics, etc).
  1. Raise an exception (current behavior)
  2. Send the packet anyway, and let the receiver deal with the problem.
  3. Silently drop, maybe log. 

  Theoretically it is possible to support all these things by having different DatagramBuilder implementations.

### Performance

The result is that the new client is much faster than the legacy singleton.

```
       Legacy client     6.414k i/100ms
          New client     9.920k i/100ms
       Legacy client     67.000k (± 1.9%) i/s -    339.942k in   5.075621s
          New client    105.344k (± 2.1%) i/s -    535.680k in   5.087400s

Comparison:
          New client:   105343.8 i/s
       Legacy client:    67000.1 i/s - 1.57x  (± 0.00) slower
```

### TODO

- [x] Do the same name and tag normalization as the legacy implementation.
- [ ] Support for providing blocks to measure durations for `measure` and `distribution`. (And maybe `histogram`.
- [ ] Support `event` and `service_check` on DogStatsD.
- [x] Add a `LogSink`
- [x] Tests for all metric types
- [x] Tests for tag and name normalization
- [x] Tests for prefix and default tags.
- [x] Tests for datagram builder classes
- [x] Tests for sink classes
- [ ] Yardoc

### DIFF
--- a/lib/statsd/instrument/client.rb
+++ b/lib/statsd/instrument/client.rb
@@ -7,6 +7,7 @@ require 'statsd/instrument/dogstatsd_datagram_builder'
 require 'statsd/instrument/null_sink'
 require 'statsd/instrument/udp_sink'
 require 'statsd/instrument/capture_sink'
+require 'statsd/instrument/log_sink'
 
 class StatsD::Instrument::Client
   attr_reader :sink, :datagram_builder_class, :prefix, :default_tags, :sampling_disabled, :default_sample_rate
--- a/lib/statsd/instrument/client.rb
+++ b/lib/statsd/instrument/client.rb
@@ -10,12 +10,11 @@ require 'statsd/instrument/capture_sink'
 require 'statsd/instrument/log_sink'
 
 class StatsD::Instrument::Client
-  attr_reader :sink, :datagram_builder_class, :prefix, :default_tags, :sampling_disabled, :default_sample_rate
+  attr_reader :sink, :datagram_builder_class, :prefix, :default_tags, :default_sample_rate
 
   def initialize(
     sink: StatsD::Instrument::NullSink.new,
     prefix: nil,
-    sampling_disabled: false,
     default_sample_rate: 1,
     default_tags: nil,
     datagram_builder_class: StatsD::Instrument::Environment.datagram_builder_class
@@ -26,7 +25,6 @@ class StatsD::Instrument::Client
     @prefix = prefix
     @default_tags = default_tags
     @default_sample_rate = default_sample_rate
-    @sampling_disabled = sampling_disabled
   end
 
   # Emits a counter metric.
@@ -42,7 +40,8 @@ class StatsD::Instrument::Client
   # @param tags [Hash, Array] (default: nil)
   # @return StatsD::Instrument::Datagram
   def increment(name, value = 1, sample_rate: nil, tags: nil)
-    sample(sample_rate, datagram_builder.c(name, value, sample_rate, tags))
+    return unless sample?(sample_rate || @default_sample_rate)
+    emit(datagram_builder.c(name, value, sample_rate, tags))
   end
 
   # Emits a timing metric.
@@ -52,7 +51,8 @@ class StatsD::Instrument::Client
   # @param tags [Hash, Array] (default: nil)
   # @return StatsD::Instrument::Datagram
   def measure(name, value = nil, sample_rate: nil, tags: nil)
-    sample(sample_rate, datagram_builder.ms(name, value, sample_rate, tags))
+    return unless sample?(sample_rate || @default_sample_rate)
+    emit(datagram_builder.ms(name, value, sample_rate, tags))
   end
 
   # Emits a gauge metric.
@@ -62,7 +62,8 @@ class StatsD::Instrument::Client
   # @param tags [Hash, Array] (default: nil)
   # @return StatsD::Instrument::Datagram
   def gauge(name, value, sample_rate: nil, tags: nil)
-    sample(sample_rate, datagram_builder.g(name, value, sample_rate, tags))
+    return unless sample?(sample_rate || @default_sample_rate)
+    emit(datagram_builder.g(name, value, sample_rate, tags))
   end
 
   # Emits a set metric, which counts unique values.
@@ -72,7 +73,8 @@ class StatsD::Instrument::Client
   # @param tags [Hash, Array] (default: nil)
   # @return StatsD::Instrument::Datagram
   def set(name, value, sample_rate: nil, tags: nil)
-    sample(sample_rate, datagram_builder.s(name, value, sample_rate, tags))
+    return unless sample?(sample_rate || @default_sample_rate)
+    emit(datagram_builder.s(name, value, sample_rate, tags))
   end
 
   # Instantiates a new StatsD client that uses the settings of the current client,
@@ -80,12 +82,11 @@ class StatsD::Instrument::Client
   def with_options(
     sink: nil,
     prefix: nil,
-    sampling_disabled: nil,
     default_sample_rate: nil,
     default_tags: nil,
     datagram_builder_class: nil
   )
-    client = clone_with_options(sink: sink, prefix: prefix, sampling_disabled: sampling_disabled,
+    client = clone_with_options(sink: sink, prefix: prefix,
       default_sample_rate: default_sample_rate, default_tags: default_tags,
       datagram_builder_class: datagram_builder_class)
 
@@ -95,7 +96,6 @@ class StatsD::Instrument::Client
   def clone_with_options(
     sink: nil,
     prefix: nil,
-    sampling_disabled: nil,
     default_sample_rate: nil,
     default_tags: nil,
     datagram_builder_class: nil
@@ -103,7 +103,6 @@ class StatsD::Instrument::Client
     self.class.new(
       sink: sink || @sink,
       prefix: prefix || @prefix,
-      sampling_disabled: sampling_disabled || @sampling_disabled,
       default_sample_rate: default_sample_rate || @default_sample_rate,
       default_tags: default_tags || @default_tags,
       datagram_builder_class: datagram_builder_class || @datagram_builder_class,
@@ -132,11 +131,12 @@ class StatsD::Instrument::Client
     @datagram_builder ||= @datagram_builder_class.new(prefix: prefix, default_tags: default_tags)
   end
 
-  def sample(sample_rate, datagram)
-    sample_rate ||= @default_sample_rate
-    if sample_rate == 1 || sampling_disabled || rand < sample_rate
-      @sink << datagram
-    end
+  def sample?(sample_rate)
+    sample_rate == 1 || rand < sample_rate
+  end
+
+  def emit(datagram)
+    @sink << datagram
     nil
   end
 end
--- a/lib/statsd/instrument/client.rb
+++ b/lib/statsd/instrument/client.rb
@@ -1,0 +1,140 @@
+# frozen_string_literal: true
+
+require 'statsd/instrument/datagram'
+require 'statsd/instrument/statsd_datagram_builder'
+require 'statsd/instrument/dogstatsd_datagram_builder'
+require 'statsd/instrument/null_sink'
+require 'statsd/instrument/udp_sink'
+require 'statsd/instrument/capture_sink'
+
+class StatsD::Instrument::Client
+  attr_reader :sink, :datagram_builder_class, :prefix, :default_tags, :sampling_disabled, :default_sample_rate
+
+  def initialize(
+    sink: StatsD::Instrument::NullSink.new,
+    prefix: nil,
+    sampling_disabled: false,
+    default_sample_rate: 1,
+    default_tags: nil,
+    datagram_builder_class: StatsD::Instrument::Environment.datagram_builder_class
+  )
+    @sink = sink
+    @datagram_builder_class = datagram_builder_class
+
+    @prefix = prefix
+    @default_tags = default_tags
+    @default_sample_rate = default_sample_rate
+    @sampling_disabled = sampling_disabled
+  end
+
+  # Emits a counter metric.
+  # @param name [String] The name of the metric.
+  # @param value [Integer] (default: 1) The value to increment the counter by.
+  #
+  #   You should not compensate for the sample rate using the counter increment. E.g., if
+  #   your sample rate is 0.01, you should <b>not</b> use 100 as increment to compensate for it.
+  #   The sample rate is part of the packet that is being sent to the server, and the server
+  #   should know how to handle it.
+  #
+  # @param sample_rate [Float] (default: nil)
+  # @param tags [Hash, Array] (default: nil)
+  # @return StatsD::Instrument::Datagram
+  def increment(name, value = 1, sample_rate: nil, tags: nil)
+    sample(sample_rate, datagram_builder.c(name, value, sample_rate, tags))
+  end
+
+  # Emits a timing metric.
+  # @param name [String] The name of the metric.
+  # @param value [Numeric] The timing to record in milliseconds
+  # @param sample_rate [Float] (default: nil)
+  # @param tags [Hash, Array] (default: nil)
+  # @return StatsD::Instrument::Datagram
+  def measure(name, value = nil, sample_rate: nil, tags: nil)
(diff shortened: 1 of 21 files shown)


lib/statsd/instrument/datagram.rb
```diff
@@ -0,0 +1,77 @@
...
+    def normalize_tags(tags)
+      return [] unless tags
+      tags = tags.map { |k, v| k.to_s + ":" + v.to_s } if tags.is_a?(Hash)
+      tags.map { |tag| tag.tr('|,', '') }
+    end
+
+    protected
+
+    def generic(name, value, type, sample_rate: nil, tags: nil)
+      datagram = new
```
casperisfine: I don't quite get why it need to be a String subclass. I get this is not finished yet, but at this stage it could perfectly return a regular string.

And if you need to parse a datagram, IMO it's better to do it explicitly by invoking a parser. It's a bit of a slipery slope argument, but I can see some code invoking `#sample_rate` or similar without realizing that it's parsing the string and generating almost 10 objects.

lib/statsd/instrument/client.rb
```diff
@@ -0,0 +1,53 @@
...
+    sample_rate = sample_rate || @default_sample_rate
+
+    datagram = datagram_builder.counter(
+      @prefix ? "#{@prefix}.#{name}" : name,
+      value,
+      sample_rate: sample_rate,
+      tags: datagram_builder.normalize_tags(tags) + default_tags,
+    )
+
+    if sample_rate == 1 || sampling_disabled || rand < sample_rate
```
casperisfine: IMO this logic move inside `emit`.

lib/statsd/instrument/datagram.rb
```diff
@@ -0,0 +1,77 @@
...
+
+
+    # Utility function to convert tags to the canonical form.
+    #
+    # - Tags specified as key value pairs will be converted into an array
+    # - Tags are normalized to only use word characters and underscores.
+    #
+    # @param tags [Array<String>, Hash<String, String>, nil] Tags specified in any form.
+    # @return [Array<String>, nil] the list of tags in canonical form.
+    def normalize_tags(tags)
```
casperisfine: The way StatsD process tags is also quite a big source of allocations. Do you think it could be the occasion to be more strict about it?

lib/statsd/instrument/datagram.rb
```diff
@@ -0,0 +1,77 @@
...
+      return [] unless tags
+      tags = tags.map { |k, v| k.to_s + ":" + v.to_s } if tags.is_a?(Hash)
```
casperisfine: `"#{k}:#{v}"`

lib/statsd/instrument/datagram.rb
wvanbergen: I considered it, but I worry about the migration path.

lib/statsd/instrument/datagram.rb
wvanbergen: > And if you need to parse a datagram

Parsing is basically never needed, *except* for testing: e.g. if you want to verify the right StatsD metric is being emitted. For that use case I am OK taking a bit of a performance hit.

We could use strings and have a separate parser class to make this more explicit. However, with this approach I think it is possible to make this object duck-type compatible with the old Metric class, which will make it a lot easier to migrate existing test suites.

casperisfine: Ok, I get it now, but then I'd simply wrap the datagram in such class when exposing it to test helpers.

wvanbergen: I have moved to using strings.

wvanbergen:
This is starting to be a lot more complete. I have implemented different sink classes, and different datagram builders to validate the approach. I've also added some basic tests and a benchmark script. I have updated the PR description to reflect these changes.

Question: What should be part of this PR, and what should be part as a followup? This new client does not interfere with the old code at all, so we can merge this PR and iterate in followup PRs if that is our preferred approach.

dylanahsmith:
> Right now, StatsD.instrument returns a StatsD::Instrument::Metric instance. The new implementation does not use this class anymore. What should we do?
> iii. Assume that basically nobody really uses the return value. Break backwards compatibility and return `void`.

iii would be my preference.  We can also have it return `nil` so we don't accidentally return something useful.  We don't really care about the return value in production code.  At the very least, we should do this in the next major release, even if we keep it temporarily for now.

> How should we handle when the user asks for something that the implementation does not support? (E.g. tags, distribution metrics, etc).

My first instinct would be to have it raise to catch the mistake.  However, perhaps it could be useful to not have it raise for library code that wants to support multiple implementations or for an application trying to transition to a different backend.

wvanbergen:
> My first instinct would be to have it raise to catch the mistake. However, perhaps it could be useful to not have it raise for library code that wants to support multiple implementations or for an application trying to transition to a different backend.

This is implemented now. We have a base DatagramBuilder class that implements everything. Specific subclasses overwrite some methods that raise a `NotImplementedError` exception

wvanbergen:
> iii would be my preference. We can also have it return nil so we don't accidentally return something useful. We don't really care about the return value in production code. At the very least, we should do this in the next major release, even if we keep it temporarily for now.

I've implemented returning `nil` explicitly for the metric methods on the new client. We won't make this client the new default until the next major version.

wvanbergen:
I think this is good to review now. There is more work to do but I rather do that in followup PRs.

wvanbergen:
With @dylanahsmith's refactors of the old client, the performance difference is now less pronounced, but definitely still significant:

```
       Legacy client     7.787k i/100ms
          New client     9.786k i/100ms
       Legacy client     82.187k (± 2.3%) i/s -    412.711k in   5.024246s
          New client    103.670k (± 2.4%) i/s -    518.658k in   5.005955s

Comparison:
          New client:   103669.9 i/s
       Legacy client:    82187.2 i/s - 1.26x  (± 0.00) slower
```

lib/statsd/instrument/datagram_builder.rb
```diff
@@ -0,0 +1,91 @@
...
+    end
+  end
+
+  def initialize(prefix: nil, default_tags: nil)
+    @prefix = prefix.nil? ? "" : "#{normalize_name(prefix)}."
+    @default_tags = normalize_tags(default_tags)
+  end
+
+  def c(name, value, sample_rate, tags)
+    generate_generic_datagram(name, value, -'c', sample_rate, tags)
```
casperisfine: I don't see the use for `String#-@` here. These are string literals inside a `frozen_string_literal: true` file, so they already are deduplicated.

Also it would break on MRI 2.3.

lib/statsd/instrument/datagram_builder.rb
```diff
@@ -0,0 +1,91 @@
+# frozen_string_literal: true
+
+class StatsD::Instrument::DatagramBuilder
+  unless Regexp.method_defined?(:match?) # for ruby 2.3
+    module RubyBackports
+      refine Regexp do
+        def match?(str)
+          (self =~ str) != nil
```
casperisfine: total nitpick, but `match(str) != nil` would trigger my OCD less, and would avoid setting the useless `$` magic variables.

Not a huge diff though.

lib/statsd/instrument/client.rb
```diff
@@ -0,0 +1,142 @@
...
+  def with_capture_sink(capture_sink)
+    @sink = capture_sink
+    yield
+    @sink.datagrams
+  ensure
+    @sink = @sink.parent
+  end
+
+  def capture(&block)
+    with_capture_sink(capture_sink, &block)
```
casperisfine: I don't quite see the value of `with_capture_sink` being extracted, `capture` end up being a one liner, and only caller of `with_capture_sink`. Seems like both method could be merged without any increase in complexity.

lib/statsd/instrument/udp_sink.rb
```diff
@@ -0,0 +1,45 @@
...
+
+  rescue SocketError, IOError, SystemCallError
+    # TODO: log?
+    invalidate_socket
+  end
+
+  private
+
+  def with_socket
+    @mutex.synchronize { yield(socket) }
```
casperisfine: Kind of note to self: it would be interesting to measure the cost of that mutex, and compare with potential alternatives (e.g. per thread socket).

lib/statsd/instrument/client.rb
wvanbergen: I can leave this out, but we need this to implement `assert_statsd_`-methods correctly. See https://github.com/Shopify/statsd-instrument/pull/166

lib/statsd/instrument/datagram_builder.rb
wvanbergen: Fixed!

lib/statsd/instrument/client.rb
```diff
@@ -0,0 +1,142 @@
...
+  # @param value [Integer] (default: 1) The value to increment the counter by.
+  #
+  #   You should not compensate for the sample rate using the counter increment. E.g., if
+  #   your sample rate is 0.01, you should <b>not</b> use 100 as increment to compensate for it.
+  #   The sample rate is part of the packet that is being sent to the server, and the server
+  #   should know how to handle it.
+  #
+  # @param sample_rate [Float] (default: nil)
+  # @param tags [Hash, Array] (default: nil)
+  # @return StatsD::Instrument::Datagram
```
dylanahsmith: looks like you forgot to update the documentation to show that we are changing to fire and forget style methods

```suggestion
  # @return [void]
```

lib/statsd/instrument/client.rb
```diff
@@ -0,0 +1,142 @@
...
+  def increment(name, value = 1, sample_rate: nil, tags: nil)
+    sample(sample_rate, datagram_builder.c(name, value, sample_rate, tags))
```
dylanahsmith: We shouldn't build the datagram if we aren't sampling this instrumentation call.

```suggestion
    return unless sample?(sample_rate)
    emit(datagram_builder.c(name, value, sample_rate, tags))
```

where the `sample` method would be replaced by `emit` and `sample?` which can be defined as follows

```ruby
  def sample?(sample_rate)
    sample_rate ||= @default_sample_rate
    sample_rate == 1 || sampling_disabled || rand < sample_rate
  end

  def emit(datagram)
    @sink << datagram
    nil
  end
```

lib/statsd/instrument/environment.rb
```diff
@@ -22,6 +22,15 @@ def default_backend
     end
   end
 
+  def datagram_builder_class
+    case ENV['STATSD_IMPLEMENTATION']
+    when 'datadog', 'dogstatsd'
+      StatsD::Instrument::DogStatsDDatagramBuilder
+    else
```
dylanahsmith: Shouldn't we also have `case 'statsite'`?  Since we still have the statsite specific `key_value` method, I assume we are still supporting it.

lib/statsd/instrument/datagram_builder.rb
```diff
@@ -0,0 +1,91 @@
...
+        end
+      end
+    end
+
+    using RubyBackports
+  end
+
+  def self.unsupported_datagram_types(*types)
+    types.each do |type|
+      define_method(type) do |_, _, _, _|
```
dylanahsmith: Instead of undefining methods in the derived classes, perhaps we should only define those methods in the derived classes.

lib/statsd/instrument/udp_sink.rb
dylanahsmith: Perhaps we should do the synchronization outside of the sync so that we can easily choose to have client per-thread.  We could have a SynchronizedSync that wraps the `UDPSink` with this mutex synchronization.

lib/statsd/instrument/client.rb
```diff
@@ -0,0 +1,142 @@
...
+require 'statsd/instrument/capture_sink'
+require 'statsd/instrument/log_sink'
+
+class StatsD::Instrument::Client
+  attr_reader :sink, :datagram_builder_class, :prefix, :default_tags, :sampling_disabled, :default_sample_rate
+
+  def initialize(
+    sink: StatsD::Instrument::NullSink.new,
+    prefix: nil,
+    sampling_disabled: false,
```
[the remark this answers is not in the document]
wvanbergen: I wasn't too sure what the best way to handle it.

If we leave it up to the sink, it needs to know the sample rate of the datagram. It can do that by parsing the string (a no go area), or we have to explicitly pass it in which requires changing the contract. I kind of liked that right now a sink only needs to implement `<<(packet)` and we can mock it with a simple array. But the alternatives are probably worse.

lib/statsd/instrument/environment.rb
wvanbergen: it should, haven't got around to implement it yet.

lib/statsd/instrument/datagram_builder.rb
wvanbergen: We don't really have a choice to expose all the possible metric types as methods on the `StatsD::Instrument::Client` class, unless we provide different subclasses. I want to make sure we are raising a `NotImplementError`, rather than a `NoMethodError` if somebody uses a metric type that is not supported by their implementation.

We can either use a whitelist approach (your proposal) or a blacklist (current implementation). I chose the blacklist approach because it was less duplication, but I don't feel super strongly about it.

lib/statsd/instrument/datagram.rb
wvanbergen: It is possible to provide a separate `DatagramBuilder` class that doesn't do normalization, and make the user responsible. I would not want to enable that by default though.

wvanbergen:
I would like to get this PR merged and do any followup work in separate PRs. Given that the code is not even loaded by default, it won't do much harm on `master`. Afterwards we can improve different aspects of this new client in parallel.

I have removed the `sampling_enabled` property for now. We still need to figure out how we want to support not sampling metrics for the capture and log sinks, even when a sample rate is set in a nice way.

Anything that you still want me to address in this PR?

lib/statsd/instrument/client.rb
wvanbergen: I've implemented this approach, but removed `sampling_disabled` for now as an issue we'll have to address in a followup (see below).

lib/statsd/instrument/datagram_builder.rb
dylanahsmith: We might want to consider client subclasses for extensions that are specific to a backend.  I think it would be completely fine for `NoMethodError` to be raised because a method is called on a client that doesn't support that method.  However, in the case of an extension like `tags` that are datadog specific, I don't see an easy way of extending the method without basically re-implementing it.  So I'm not sure if it is worth doing.

lib/statsd/instrument/datagram.rb
```diff
@@ -0,0 +1,53 @@
...
+  def name
+    parsed_datagram[:name]
+  end
+
+  def value
+    parsed_datagram[:value]
+  end
+
+  def tags
+    parsed_datagram[:tags] ? parsed_datagram[:tags].split(',') : nil
```
dylanahsmith: This could be memoized

PR #173, opened 2019-09-23
Use match? on the fast path for normalization

dylanahsmith:
Regexp `match?` avoids having to do any object allocations by just having to return a boolean result.  `Array#any?` avoids object allocations for the same reason.

This improvement shows up in the benchmark:

```
$ ./benchmark/send-metrics-to-local-udp-receiver
StatsD metrics to local UDP receiver (branch: normalization-fast-path, sha: db42a9d)
                       926.000  i/100ms
StatsD metrics to local UDP receiver (branch: normalization-fast-path, sha: db42a9d)
                          9.376k (± 2.3%) i/s -     47.226k in   5.039640s

Comparison:
StatsD metrics to local UDP receiver (branch: normalization-fast-path, sha: db42a9d):     9376.2 i/s
StatsD metrics to local UDP receiver (branch: master, sha: 779193d):     8857.1 i/s - same-ish: difference falls within error
```

Although is based on the assumption that metric names don't have `:`, `|` or `@` in them and tag names don't have `|` or`,`.  When these assumptions aren't true, then these checks could make things a bit slower.

### DIFF
--- a/lib/statsd/instrument/metric.rb
+++ b/lib/statsd/instrument/metric.rb
@@ -123,6 +123,9 @@ class StatsD::Instrument::Metric
   # @param name [String]
   # @return [String]
   def normalize_name(name)
+    # fast path when no normalization is needed to avoid copying the string
+    return name unless name.match?(/[:|@]/)
+
     name.tr(':|@', '_')
   end
 
@@ -136,6 +139,10 @@ class StatsD::Instrument::Metric
   def self.normalize_tags(tags)
     return unless tags
     tags = tags.map { |k, v| k.to_s + ":" + v.to_s } if tags.is_a?(Hash)
+
+    # fast path when no string replacement is needed
+    return tags unless tags.any? { |tag| tag.match?(/[|,]/) }
+
     tags.map { |tag| tag.tr('|,', '') }
   end
 end
--- a/lib/statsd/instrument/metric.rb
+++ b/lib/statsd/instrument/metric.rb
@@ -31,6 +31,18 @@
 # @see StatsD::Instrument::Backend A StatsD::Instrument::Backend is used to collect metrics.
 #
 class StatsD::Instrument::Metric
+  unless Regexp.method_defined?(:match?) # for ruby 2.3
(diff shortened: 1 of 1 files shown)


wvanbergen:
I maintained 2.3 support because that's the version that comes with macOS.

casperisfine:
It's also possible to use `String.method_defined?(:match?)` to define two alternative versions of these methods.

dylanahsmith:
I can use a refinement on ruby 2.3 to backport that locally in a way that is easy to remove once we drop support for the EOL ruby 2.3.

lib/statsd/instrument/metric.rb
```diff
@@ -31,6 +31,18 @@
 # @see StatsD::Instrument::Backend A StatsD::Instrument::Backend is used to collect metrics.
 #
 class StatsD::Instrument::Metric
+  if Gem::Version.new(RUBY_VERSION) < Gem::Version.new('2.4')
```
casperisfine: I'd rather use `Regexp.method_defined?(:match?)` as a condition, because RUBY_VERSION check can be tricky with alternative implementations.

wvanbergen:
By the way, it looks like one of the tests (I assume the one that forks) times out the CI run if something goes wrong. I'll see if I can fix that.

lib/statsd/instrument/metric.rb
dylanahsmith: Mostly I just wanted to make it clear when this could be removed, but I could do feature detection and us a comment for that.  However, we should probably add any alternative ruby versions we want to support to CI (e.g. truffleruby).  I would actually be more concerned with refinements, which would be more complex to implement, than with `Regexp#match?`, which is easy to implement.

casperisfine: Yeah but that's my point. I know that jruby / truffle / etc have `Regexp#match?` however they're likely to return something you don't expect in `RUBY_VERSION`.

So I'm worried you might be using a useless refinement on those.

But yeah 👍 on adding them to CI.

PR #174, opened 2019-09-23
Use keyword args to capture hash options to avoid hash allocation

dylanahsmith:
There were a bunch of objects being allocated to handle keyword-like operations, which was being handled as a hash, as well as converting positional arguments to keyword arguments.  Using keyword argument directly allows ruby to handle the arguments without them being converted to a hash as well as letting us leverage default values.

The benchmark results show the improvement:

```
$ benchmark/send-metrics-to-local-udp-receiver
StatsD metrics to local UDP receiver (branch: master, sha: 779193d)
                       903.000  i/100ms
StatsD metrics to local UDP receiver (branch: master, sha: 779193d)
                          9.221k (± 1.8%) i/s -     46.956k in   5.093885s

Comparison:
StatsD metrics to local UDP receiver (branch: keyword-args, sha: 7611487):    11026.8 i/s
StatsD metrics to local UDP receiver (branch: master, sha: 779193d):     9221.2 i/s - 1.20x  (± 0.00) slower
```

### DIFF
--- a/lib/statsd/instrument.rb
+++ b/lib/statsd/instrument.rb
@@ -346,11 +346,17 @@ module StatsD
   #      http_response = StatsD.measure('HTTP.call.duration') do
   #        HTTP.get(url)
   #      end
-  def measure(key, value = nil, *metric_options, &block)
-    value, metric_options = parse_options(value, metric_options)
-    type = (!metric_options.empty? && metric_options.first[:as_dist] ? :d : :ms)
-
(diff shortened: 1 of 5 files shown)


casperisfine:
You'll have to appease rubocop though.

wvanbergen:
Alternatively, additionally to this: could we write a Rubocop rule to rewrite all the StatsD calls to follow the canonical `StatsD.increment(name, value, keyword:, arguments:)`?

PR #176, opened 2019-09-23
Add custom Rubocop cop to rewrite positional arguments on StatsD calls.

wvanbergen:
This rewrites deprecated positional arguments for `StatsD.metric` calls to use keyword arguments instead.

Currently, `value` can also be provides as keyword argument rather than positional argument, which is not the preferred way of using this. The cop does not correct that (yet).

### DIFF
--- a/lib/statsd/instrument/rubocop/positional_arguments.rb
+++ b/lib/statsd/instrument/rubocop/positional_arguments.rb
@@ -7,16 +7,25 @@ module RuboCop
         MSG = 'Use keyword arguments for StatsD calls'
 
         STATSD_SINGLETON_METHODS = %i{increment gauge measure set histogram distribution key_value}
+
         POSITIONAL_ARGUMENT_TYPES = Set[:int, :float, :nil]
-        UNKNOWN_ARGUMENT_TYPES = Set[:send, :const]
+        UNKNOWN_ARGUMENT_TYPES = Set[:send, :const, :lvar]
         REFUSED_ARGUMENT_TYPES = POSITIONAL_ARGUMENT_TYPES | UNKNOWN_ARGUMENT_TYPES
+
+        KEYWORD_ARGUMENT_TYPES = Set[:hash]
+        BLOCK_ARGUMENT_TYPES = Set[:block_pass]
+        ACCEPTED_ARGUMENT_TYPES = KEYWORD_ARGUMENT_TYPES | BLOCK_ARGUMENT_TYPES
 
         def on_send(node)
           if node.receiver&.type == :const && node.receiver&.const_name == "StatsD"
-            if STATSD_SINGLETON_METHODS.include?(node.method_name)
-              arguments = node.arguments
-              if arguments.length >= 3 && REFUSED_ARGUMENT_TYPES.include?(arguments[2].type)
+            if STATSD_SINGLETON_METHODS.include?(node.method_name) && node.arguments.length >= 3
+              case node.arguments[2].type
+              when *REFUSED_ARGUMENT_TYPES
                 add_offense(node)
+              when *ACCEPTED_ARGUMENT_TYPES
+                nil
+              else
+                $stderr.puts "[StatsD/PositionalArguments] Unhandled argument type: #{node.arguments[2].type.inspect}"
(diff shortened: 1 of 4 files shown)


lib/statsd/instrument/rubocop/positional_arguments.rb
```diff
@@ -0,0 +1,59 @@
...
+      class PositionalArguments < Cop
+        MSG = 'Use keyword arguments for StatsD calls'
+
+        STATSD_SINGLETON_METHODS = %i{increment gauge measure set histogram distribution key_value}
+
+        def on_send(node)
+          if node.receiver&.type == :const && node.receiver&.const_name == "StatsD"
+            if STATSD_SINGLETON_METHODS.include?(node.method_name)
+              arguments = node.arguments
+              if arguments.length > 2 && ![:block_pass, :hash].include?(arguments[2].type)
```
dylanahsmith: Should `[:block_pass, :hash]` also be moved to a constant to avoid allocation each time this line is run?

test/rubocop/positional_arguments_test.rb
```diff
@@ -0,0 +1,112 @@
...
+      corrected = autocorrect_source("StatsD.measure('foo', 2, nil, foo: 'bar', &block)")
+      assert_equal "StatsD.measure('foo', 2, tags: { foo: 'bar' }, &block)", corrected
+    end
+
+    def test_autocorrect_sample_rate_and_curly_braces_block
+      corrected = autocorrect_source("StatsD.measure('foo', 2, 0.5) { foo }")
+      assert_equal "StatsD.measure('foo', 2, sample_rate: 0.5) { foo }", corrected
+    end
+
+    def test_autocorrect_sample_rate_and_d0_end_block
```
dylanahsmith:
```suggestion
    def test_autocorrect_sample_rate_and_do_end_block
```

lib/statsd/instrument/rubocop/positional_arguments.rb
```diff
@@ -0,0 +1,60 @@
...
+        MSG = 'Use keyword arguments for StatsD calls'
+
+        STATSD_SINGLETON_METHODS = %i{increment gauge measure set histogram distribution key_value}
+        NON_POSITIONAL_SECOND_ARGUMENT_TYPES = [:block_pass, :hash]
+
+        def on_send(node)
+          if node.receiver&.type == :const && node.receiver&.const_name == "StatsD"
+            if STATSD_SINGLETON_METHODS.include?(node.method_name)
+              arguments = node.arguments
+              if arguments.length > 2 && !NON_POSITIONAL_SECOND_ARGUMENT_TYPES.include?(arguments[2].type)
```
dylanahsmith: What if the options hash is passed using a local variable or through a keyword splat (e.g. `**kwargs`)?

wvanbergen: I've been testing this cop on Shopify/shopify. 

`**kwargs` seem to work fine.

However, when a method is provided, the result is ambiguous, because we don't know whether this method will return a hash (interpreted as keyword arguments) or a sample rate.

This means we can detect these cases, but we'll need to manually rewrite them.

lib/statsd/instrument/rubocop/positional_arguments.rb
```diff
@@ -0,0 +1,90 @@
...
+        def on_send(node)
+          if node.receiver&.type == :const && node.receiver&.const_name == "StatsD"
+            if STATSD_SINGLETON_METHODS.include?(node.method_name) && node.arguments.length >= 3
+              case node.arguments[2].type
+              when *REFUSED_ARGUMENT_TYPES
+                add_offense(node)
+              when *ACCEPTED_ARGUMENT_TYPES
+                nil
+              else
+                $stderr.puts "[StatsD/PositionalArguments] Unhandled argument type: #{node.arguments[2].type.inspect}"
```
wvanbergen: This never gets triggered in Shopify/shopify

wvanbergen:
I've reworked this to be a bit more careful: in some cases it will find offenses, but won't autocorrect.

PR #179, opened 2019-09-25
Add a custom Rubocop cop that checks for using `value` as keyword argument

wvanbergen:
Using `StatsD.increment 'foo, value: 3` is deprecated - you should use `StatsD.increment('foo', 3)` instead. This rule will check for violations as best as it can.

I ran this cop on Shopify/shopify, and found no violations.

### DIFF
--- a/lib/statsd/instrument/rubocop/metric_value_keyword_argument.rb
+++ b/lib/statsd/instrument/rubocop/metric_value_keyword_argument.rb
@@ -23,9 +23,13 @@ module RuboCop
         def on_send(node)
           if node.receiver&.type == :const && node.receiver&.const_name == "StatsD"
             if STATSD_METRIC_METHODS.include?(node.method_name)
-              arguments = node.arguments
-              arguments.pop if node.arguments.last.type == :block_pass
-              check_keyword_arguments_for_value_entry(node, arguments.pop) if arguments.last&.type == :hash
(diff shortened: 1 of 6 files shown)


lib/statsd/instrument/rubocop/metric_value_keyword_argument.rb
```diff
@@ -0,0 +1,42 @@
...
+      class MetricValueKeywordArgument < Cop
+        MSG = 'Do not use the value keyword argument, but use a positional argument'
+
+        STATSD_METRIC_METHODS = %i{increment gauge measure set histogram distribution key_value}
+
+        def on_send(node)
+          if node.receiver&.type == :const && node.receiver&.const_name == "StatsD"
+            if STATSD_METRIC_METHODS.include?(node.method_name)
+              arguments = node.arguments
+              arguments.pop if node.arguments.last.type == :block_pass
```
dylanahsmith: Is it safe to mutate `node.arguments`?  Couldn't that affect other cops?

wvanbergen: I'll add a dup just in case.

wvanbergen: Rather than dupping I changed it to simply extract the last argument into a variable.

PR #181, opened 2019-09-25
Rubocop: run our own cops on this repo.

wvanbergen:
We have some nice Rubocop rules. We should make sure our codebase conforms. By doing so, I already found some interesting things:

- `StatsD.key_value`'s third argument is secretly actually a timestamp, not a sample rate. So changing this to `sample_rate: timestamp` does not make sense. We probably should figure out how we want to handle this. The README doesn't mention this at all (https://github.com/statsite/statsite), so maybe we should simply :fire: this?

- Our metaprogramming methods (`stats_count` and friends) still rely on positional arguments. I am pretty sure I also found a bug (a missing value argument). How do we want to deal with this? I am not a big fan of these methods, but we cannot really get rid of them.

### DIFF
--- a/lib/statsd/instrument.rb
+++ b/lib/statsd/instrument.rb
@@ -84,6 +84,8 @@ module StatsD
       Process.clock_gettime(Process::CLOCK_MONOTONIC) - start
     end
 
+    # rubocop:disable StatsD/PositionalArguments
+
     # Adds execution duration instrumentation to a method as a timing.
     #
     # @param method [Symbol] The name of the method to instrument.
@@ -191,6 +193,7 @@ module StatsD
           ensure
             if truthiness
               metric_name = StatsD::Instrument.generate_metric_name(name, self, *args)
+              # TODO: is this a bug, because 1 is missing?
               StatsD.increment(metric_name, *metric_options)
             end
           end
@@ -216,6 +219,8 @@ module StatsD
         end
       end
     end
+
+    # rubocop:enable StatsD/PositionalArguments
 
     # Removes StatsD counter instrumentation from a method
     # @param method [Symbol] The method to remove instrumentation from.
--- a/test/udp_backend_test.rb
+++ b/test/udp_backend_test.rb
@@ -148,11 +148,19 @@ class UDPBackendTest < Minitest::Test
     StatsD.key_value('fooy', 42)
   end
 
+  # For key_value metrics (only supported by statsite), the sample rate
+  # part of the datagram format is (ab)used to be set to a timestamp instead.
+  # Changing that to `sample_rate: timestamp` does not make sense, so we
+  # disable the rubocop rule for positional arguments for now,
+  # until we figure out how we want to handle this.
+
+  # rubocop:disable StatsD/PositionalArguments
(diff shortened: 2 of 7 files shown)


dylanahsmith:
> * `StatsD.key_value`'s third argument is secretly actually a timestamp, not a sample rate.

How did you find out about that?  Is the timestamp an expiration time?  If we want to support that, then we should use an appropriately named keyword argument.

> * Our metaprogramming methods (`stats_count` and friends) still rely on positional arguments.

You mean the usage of them in the application?  Since it seems like they just pass options with a `*metric_options` splat which supports both keyword and positional arguments.

lib/statsd/instrument.rb
```diff
@@ -191,6 +193,7 @@ def statsd_count_if(method, name, *metric_options)
           ensure
             if truthiness
               metric_name = StatsD::Instrument.generate_metric_name(name, self, *args)
+              # TODO: is this a bug, because 1 is missing?
```
dylanahsmith: It is probably a bug, given that it is inconsistent with `statsd_count_success` in this way.  That means that the `value` is being provided in the third positional argument to `statsd_count_if`, but in `statsd_count_success` it is the sample rate.

test/udp_backend_test.rb
```diff
@@ -148,11 +148,19 @@ def test_supports_key_value_syntax_on_statsite
...
+  # part of the datagram format is (ab)used to be set to a timestamp instead.
+  # Changing that to `sample_rate: timestamp` does not make sense, so we
+  # disable the rubocop rule for positional arguments for now,
+  # until we figure out how we want to handle this.
+
+  # rubocop:disable StatsD/PositionalArguments
   def test_supports_key_value_with_timestamp_on_statsite
     @backend.implementation = :statsite
     @backend.expects(:write_packet).with("fooy:42|kv|@123456\n")
     StatsD.key_value('fooy', 42, 123456)
```
dylanahsmith: So all we really know about this timestamp parameter is from this test?

wvanbergen: Yeah - Statsite support was contributed at some point - we've never used it ourselves. So all we have are the tests. The statsite repo has a bit more info though.

lib/statsd/instrument.rb
wvanbergen: Another good reason to not use positional arguments :)

I fixed this in https://github.com/Shopify/statsd-instrument/pull/182/

wvanbergen:
> You mean the usage of them in the application? Since it seems like they just pass options with a `*metric_options` splat which supports both keyword and positional arguments

This works fine (except in the case where we have the bug), but it doesn't satisfy our Rubocop rule about positional arguments.

Now I could also change the Rubocop rule, but I decided against it. After `StatsD.metric(name, value, ...` we should expect only keyword arguments. So I think it is correct for the cop to tell people to change it to `**metric_options` instead. The rule will not auto-correct this though.

Either way, I fixed this issue in #182.

wvanbergen:
Anyway, all the issues I brought up should be fixed in separate PRs - the focus of this PR is simply to get Rubocop to run with our own rules on our repo.

PR #182, opened 2019-09-25
Fix statsd_* metaprogramming reliance on positional arguments.

wvanbergen:
Ref https://github.com/Shopify/statsd-instrument/pull/181

This fixes the `statsd_*` metaprogramming to accept both positional as well as keyword arguments, following the approach of #174. When calling the underlying StatsD metric method, we use keyword arguments so we pass our own Rubocop rules

### DIFF
--- a/lib/statsd/instrument.rb
+++ b/lib/statsd/instrument.rb
@@ -84,8 +84,6 @@ module StatsD
       Process.clock_gettime(Process::CLOCK_MONOTONIC) - start
     end
 
-    # rubocop:disable StatsD/PositionalArguments
-
     # Adds execution duration instrumentation to a method as a timing.
     #
     # @param method [Symbol] The name of the method to instrument.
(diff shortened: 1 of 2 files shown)


lib/statsd/instrument.rb
```diff
@@ -94,8 +94,8 @@ def self.duration
     def statsd_measure(method, name, *metric_options)
       add_to_method(method, name, :measure) do
         define_method(method) do |*args, &block|
-          metric_name = StatsD::Instrument.generate_metric_name(name, self, *args)
-          StatsD.measure(metric_name, *metric_options) { super(*args, &block) }
+          key = StatsD::Instrument.generate_metric_name(name, self, *args)
+          StatsD.measure(key, *metric_options) { super(*args, &block) }
```
dylanahsmith: Why are we fine the the args splat here, but avoid doing that in `statsd_distribution`?

wvanbergen: Ugh, I missed this.

wvanbergen: This is why we need #181 :P

PR #183, opened 2019-09-25
Add a Strict monkeypatch to find deprecated behavior

wvanbergen:
Adds a `Strict` monkeypatch that strictly validates the arguments provided to StatsD methods for correct usage. You can enable this in CI of a project by adding `require 'statsd/instrument/strict'`, to fin deprecated behavior that may no longer work in the future.

### DIFF
--- a/lib/statsd/instrument/strict.rb
+++ b/lib/statsd/instrument/strict.rb
@@ -1,0 +1,103 @@
+# frozen_string_literal: true
+
+require 'statsd-instrument' unless Object.const_defined?(:StatsD)
+
+module StatsD
+  module Instrument
+    UNSPECIFIED = Object.new.freeze
+    private_constant :UNSPECIFIED
+
+    # The Strict monkeypatch can be loaded to see if you're using the StatsD library in
+    # a deprecated way.
+    #
+    # - The metric methods are not retuning a Metric instance.
+    # - Only accept keyword arguments for tags and sample_rate, rather than position arguments.
+    # - Only accept a position argument for value, rather than a keyword argument.
+    # - The provided arguments have the right type.
+    #
+    # You can enable thois monkeypatch by changing your Gemfile as follows:
+    #
+    #     gem 'statsd-instrument', require: 'statsd/instrument/strict'
+    #
+    # By doing this as part of your QA/CI, you can find where you are still using deprecated patterns,
+    # and fix them before the deprecated behavior is removed in the next major version.
+    #
+    # This monkeypatch is not meant to be used in production.
+    module Strict
+      def increment(key, value = 1, sample_rate: nil, tags: nil, prefix: StatsD.prefix, no_prefix: false)
+        raise ArgumentError, "StatsD.increment does not accept a block" if block_given?
+        raise ArgumentError, "The value argument should be an integer, got #{value.inspect}" unless value.is_a?(Numeric)
+        check_tags_and_sample_rate(sample_rate, tags)
+
+        super
+      end
+
+      def gauge(key, value, sample_rate: nil, tags: nil, prefix: StatsD.prefix, no_prefix: false)
+        raise ArgumentError, "StatsD.increment does not accept a block" if block_given?
+        raise ArgumentError, "The value argument should be an integer, got #{value.inspect}" unless value.is_a?(Numeric)
+        check_tags_and_sample_rate(sample_rate, tags)
+
+        super
+      end
+
+      def histogram(key, value, sample_rate: nil, tags: nil, prefix: StatsD.prefix, no_prefix: false)
+        raise ArgumentError, "StatsD.increment does not accept a block" if block_given?
+        raise ArgumentError, "The value argument should be an integer, got #{value.inspect}" unless value.is_a?(Numeric)
+        check_tags_and_sample_rate(sample_rate, tags)
+
+        super
+      end
+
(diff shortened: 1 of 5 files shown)


wvanbergen:
This is definitely not ready, but some early feedback would be welcome!

lib/statsd/instrument.rb
```diff
@@ -523,7 +523,7 @@ def service_check(
     collect_metric(:_sc, name, status, sample_rate: sample_rate, tags: tags, prefix: prefix, metadata: metadata)
   end
 
-  private
+  protected
```
dylanahsmith: Why was this change needed?  `protected` doesn't seem like it makes sense on a singleton class.

lib/statsd/instrument/strict.rb
```diff
@@ -0,0 +1,84 @@
...
+    # By doing this as part of your QA/CI, you can find where you are still using deprecated patterns,
+    # and fix them before the deprecated behavior is removed in the next major version.
+    #
+    # This monkeypatch is not meant to be used in production.
+    module Strict
+      def increment(key, value = 1, sample_rate: nil, tags: nil, prefix: StatsD.prefix, no_prefix: false)
+        raise ArgumentError, "StatsD.increment does not accept a block" if block_given?
+        raise ArgumentError, "The value argument should be an integer" unless value.is_a?(Integer)
+        raise ArgumentError, "The sample_rate argument should be a number" unless sample_rate.nil? || sample_rate.is_a?(Numeric)
+        raise ArgumentError, "The tags argument should be a hash or an array" unless tags.nil? || tags.is_a?(Hash) || tags.is_a?(Array)
```
dylanahsmith: We could pull the `sample_rate` and `tags` checks into a common method.

lib/statsd/instrument/strict.rb
```diff
@@ -0,0 +1,84 @@
...
+
+      def set(key, value, sample_rate: nil, tags: nil, prefix: StatsD.prefix, no_prefix: false)
+        raise ArgumentError, "The sample_rate argument should be a number" unless sample_rate.nil? || sample_rate.is_a?(Numeric)
+        raise ArgumentError, "The tags argument should be a hash or an array" unless tags.nil? || tags.is_a?(Hash) || tags.is_a?(Array)
+        super
+      end
+
+      def measure(key, value = nil, sample_rate: nil, tags: nil, prefix: StatsD.prefix, no_prefix: false, as_dist: false, &block)
+        if block_given?
+          raise ArgumentError, "The value argument should be set to nil when providing a block" unless value.nil?
```
dylanahsmith: From the library user's perspective, they should just think that they value shouldn't be given when calling it with a block.

```suggestion
          raise ArgumentError, "A value argument should be given when providing a block" unless value.nil?
```

lib/statsd/instrument/version.rb
```diff
@@ -2,6 +2,6 @@
 
 module StatsD
   module Instrument
-    VERSION = "2.4.0"
+    VERSION = "2.3.9"
```
dylanahsmith: You shouldn't have to lie anymore in Shopify core now that I've resolve the version conflict.

test/test_helper.rb
```diff
@@ -11,4 +11,12 @@
 
 require_relative 'helpers/rubocop_helper'
 
+# Uncomment this line to enable strict mode on the librray itself.
```
dylanahsmith:
```suggestion
# Uncomment this line to enable strict mode on the library itself.
```

You could use an environment variable.  That could also be used to run CI both with and without strict mode by providing the environment variable to one of them.

wvanbergen: Great idea

dylanahsmith:
Should we use sorbet for type checking?  I think we can even have the sorbet rbi file get generated from the documentation to provide that type checking.  We should be able to use the sorbet runtime during testing and not use it for production (using sorbet stubs to ignore any inline annotations in the application).

lib/statsd/instrument.rb
wvanbergen: This was left over from an unsuccessful debugging attempt. It is not needed.

test/test_helper.rb
```diff
@@ -11,4 +11,11 @@
 
 require_relative 'helpers/rubocop_helper'
 
+require 'statsd/instrument/strict' if ENV['STATSD_STRICT_MODE']
+
+def strict_mode_enabled?
```
casperisfine: Not sure i'm fan of having this defined as a main method. Can't it be inside some test helper?

wvanbergen: Fixed

wvanbergen:
So this is working pretty well, and has helped me find many issues in our StatsD usage.

What is still missing is s strict monkeypatch for the metaprgramming methods. I rather add that in a separate PR, so we can review and merge this one.

lib/statsd/instrument/strict.rb
```diff
@@ -0,0 +1,103 @@
...
+      def distribution(key, value = UNSPECIFIED, sample_rate: nil, tags: nil,
+        prefix: StatsD.prefix, no_prefix: false, &block)
+
+        check_block_or_numeric_value(value, &block)
+        check_tags_and_sample_rate(sample_rate, tags)
+
+        super
+      end
+
+      protected
```
dylanahsmith: Looks like this also could be `private`

```suggestion
      private
```

PR #186, opened 2019-09-26
Strict module for StatsD meta-programming methods

wvanbergen:
Followup to #183.

This adds a strict definition of the metaprogramming methods as part of strict mode. This can be used to find deprecated patterns in codebases that use this library.

### DIFF
--- a/lib/statsd/instrument/strict.rb
+++ b/lib/statsd/instrument/strict.rb
@@ -97,7 +97,49 @@ module StatsD
         nil # We explicitly discard the return value, so people cannot depend on it.
       end
     end
+
+    module StrictMetaprogramming
+      def statsd_measure(method, name, sample_rate: nil, tags: nil,
+        prefix: StatsD.prefix, no_prefix: false, as_dist: false)
+
+        check_method_and_metric_name(method, name)
+        super
+      end
+
+      def statsd_distribution(method, name, sample_rate: nil, tags: nil, prefix: StatsD.prefix, no_prefix: false)
+        check_method_and_metric_name(method, name)
+        super
+      end
+
+      def statsd_count_success(method, name, sample_rate: nil, tags: nil, prefix: StatsD.prefix, no_prefix: false)
+        check_method_and_metric_name(method, name)
+        super
+      end
+
+      def statsd_count_if(method, name, sample_rate: nil, tags: nil, prefix: StatsD.prefix, no_prefix: false)
+        check_method_and_metric_name(method, name)
+        super
+      end
+
+      def statsd_count(method, name, sample_rate: nil, tags: nil, prefix: StatsD.prefix, no_prefix: false)
+        check_method_and_metric_name(method, name)
+        super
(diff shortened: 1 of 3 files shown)


wvanbergen:
I want to test this on Shopify first to see if it gives useful feedback, but it's good for some initial eyes.

lib/statsd/instrument/strict.rb
```diff
@@ -97,7 +97,47 @@ def collect_metric(type, name, value, sample_rate:, tags: nil, prefix:, metadata
         nil # We explicitly discard the return value, so people cannot depend on it.
       end
     end
+
+    module StrictMetaprgramming
```
casperisfine: typo `StrictMetaprgramming/StrictMetaprogramming `

lib/statsd/instrument/strict.rb
```diff
@@ -97,7 +97,47 @@ def collect_metric(type, name, value, sample_rate:, tags: nil, prefix:, metadata
...
+
+      def statsd_count(method, name, sample_rate: nil, tags: nil, prefix: StatsD.prefix, no_prefix: false)
+        check_method_and_metric_name(method, name)
+        super
+      end
+
+      private
+
+      def check_method_and_metric_name(method, metric_name)
+        raise ArgumentError, "The method name should be provided as symbol" unless method.is_a?(Symbol)
```
casperisfine: Could be interested to say what you got instead, might make debugging much easier some times.

test/deprecations_test.rb
```diff
@@ -3,10 +3,12 @@
 require 'test_helper'
```
dylanahsmith: Can we just `return if StatsD::Instrument.strict_mode_enabled?` here to avoid multiple checks later on?

lib/statsd/instrument/strict.rb
```diff
@@ -97,7 +97,47 @@ def collect_metric(type, name, value, sample_rate:, tags: nil, prefix:, metadata
...
+        unless metric_name.is_a?(String) || metric_name.is_a?(Proc)
+          raise ArgumentError, "The metric name should be a proc or string"
+        end
+      end
+    end
   end
 end
 
 StatsD.singleton_class.prepend(StatsD::Instrument::Strict)
+StatsD::Instrument.prepend(StatsD::Instrument::StrictMetaprgramming)
```
dylanahsmith:
```suggestion
StatsD::Instrument.prepend(StatsD::Instrument::StrictMetaprogramming)
```

test/deprecations_test.rb
wvanbergen: That would work, but I kind of like it explicit like this. The test suite runs in 0.1 seconds on my machine, so over-optimizing this is not really needed.

wvanbergen:
Just going to run this on Shopify core to see if we get useful feedback

wvanbergen:
Found some issues using this, so it seems to work!

PR #187, opened 2019-09-26
StatsD/MetaprogrammingPositionalArguments cop

wvanbergen:
This cop will check for basic cases of deprecated use of positional arguments, like `statsd_measure` and `statsd_count_if`.


wvanbergen:
I ran this cop on Shopify's codebase and found no more violations.

PR #189, opened 2019-09-27
Improve documentation on the StatsD metric methods

wvanbergen:
The method signature is pretty awful because of backwards compatibility, and we don't really want to expose the deprecated options to the end user. By using the `@!method` macro, we can override the signature from Yard's point of view.

I have decided not to document a couple of keyword arguments because I may want to drop support of these in the new version:
- `prefix`
- `no_prefix`
- `as_dist` (I think this might be an internal only thing)

It turns out that many of the arguments are for the datadog-only `event` and `service_check` metrics are not actually used. To start, I have decided to simply hide all of them. In the new implementation, we should remove them from the method signature.

I've also added some examples, and properly documented the methods that optionally accept a block (`measure` and `distribution`)

### DIFF
--- a/lib/statsd/instrument.rb
+++ b/lib/statsd/instrument.rb
@@ -29,13 +29,12 @@ require 'logger'
 # @!attribute logger
 #   The logger to use in case of any errors. The logger is also used as default logger
 #   for the LoggerBackend (although this can be overwritten).
+#   @see StatsD::Instrument::Backends::LoggerBackend
+#   @return [Logger]
 #
 # @!attribute default_tags
 #   The tags to apply to all metrics.
 #   @return [Array<String>, Hash<String, String>, nil] The default tags, or <tt>nil</tt> when no default tags is used
-#
-#   @see StatsD::Instrument::Backends::LoggerBackend
-#   @return [Logger]
 #
 # @see StatsD::Instrument <tt>StatsD::Instrument</tt> contains module to instrument
 #    existing methods with StatsD metrics.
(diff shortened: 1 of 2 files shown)


lib/statsd/instrument.rb
```diff
@@ -336,31 +335,38 @@ def backend
     @backend ||= StatsD::Instrument::Environment.default_backend
   end
 
-  # Emits a duration metric.
+  # @!method measure(name, value = nil, sample_rate: nil, tags: nil, &block)
```
dylanahsmith:
```suggestion
  # @!method measure(name, sample_rate: nil, tags: nil, &block)
```

if it is given a block, then it shouldn't be given a value as well

lib/statsd/instrument.rb
```diff
@@ -336,31 +335,38 @@ def backend
...
   #
-  # @overload measure(key, metric_options = {}, &block)
-  #   Emits a measure metric, after measuring the execution duration of the
+  # @example Providing a value directly
+  #    start = Process.clock_gettime(Process::CLOCK_MONOTONIC, :millisecond)
+  #    do_something
+  #    stop = Process.clock_gettime(Process::CLOCK_MONOTONIC, :millisecond)
+  #    http_response = StatsD.measure('HTTP.call.duration', stop - start)
+  #
+  # @example Providing a block the measure the duration
```
dylanahsmith:
```suggestion
  # @example Providing a block to measure the duration of its execution
```

lib/statsd/instrument.rb
```diff
@@ -392,7 +400,17 @@ def measure(
   #   The sample rate is part of the packet that is being sent to the server, and the server
   #   should know how to handle it.
   #
-  # @param metric_options [Hash] (default: {}) Metric options
+  # @param sample_rate [Float] (default: `StatsD.default_sample_rate`) The rate at which to sample
+  #   this metric call. This value should be between 0 and 1. This value can be used to reduce
+  #   the amount of network I/O (and CPU cycles) is being used for very frequent metrics.
```
dylanahsmith:
```suggestion
  #   the amount of network I/O (and CPU cycles) used for very frequent metrics.
```

lib/statsd/instrument.rb
```diff
@@ -336,31 +335,38 @@ def backend
     @backend ||= StatsD::Instrument::Environment.default_backend
   end
 
-  # Emits a duration metric.
+  # @!method measure(name, value = nil, sample_rate: nil, tags: nil, &block)
+  # @!method measure(name, value, sample_rate: nil, tags: nil)
```
dylanahsmith: Is this macro actually meant to specify overloads?  Given that you are using `@overloads` below.  Or will this second method macro replace the first?

wvanbergen: After doing some experimentation with overloaded methods:

- The `@!method` macro ends up in the method summary list, so we should only specify one: 
Screen Shot 2019-09-27 at 2 56 14 PM

- The `@overload` signatures do not show up in the method summary list, but will show up under the method details:

Screen Shot 2019-09-27 at 2 55 41 PM

I will update it so we only use `@!method` once for the overloaded metric.

PR #190, opened 2019-09-27
Benchmark script: respect STATSD_SAMPLE_RATE.

wvanbergen:
This allows us to compare performance at different sample rates.

### DIFF
--- a/benchmark/datagram-client
+++ b/benchmark/datagram-client
@@ -23,15 +23,15 @@ new_client_receiver = UDPSocket.new
 new_client_receiver.bind('localhost', 0)
 
 udp_sink = StatsD::Instrument::UDPSink.new(new_client_receiver.addr[2], new_client_receiver.addr[1])
-new_client = StatsD::Instrument::Client.new(sink: udp_sink)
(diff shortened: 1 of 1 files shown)


benchmark/datagram-client
```diff
@@ -23,15 +23,15 @@ new_client_receiver = UDPSocket.new
 new_client_receiver.bind('localhost', 0)
 
 udp_sink = StatsD::Instrument::UDPSink.new(new_client_receiver.addr[2], new_client_receiver.addr[1])
-new_client = StatsD::Instrument::Client.new(sink: udp_sink)
+new_client = StatsD::Instrument::Client.new(sink: udp_sink, default_sample_rate: StatsD.default_sample_rate)
```
dylanahsmith: Should `default_sample_rate: StatsD.default_sample_rate` be the default for the client?

wvanbergen: Maybe/Probably? I haven't implemented it yet because I wasn't sure if I wanted to maintain this singleton setting, or get rid of it.

PR #192, opened 2019-09-30
Add DogStatsD compatibility test.

wvanbergen:
I am not sure if this is useful (and it was surprisingly hard to write, but I wrote this to affirm that the old and new implementation are compatible.

I am not so worried about the datagram format specifically, but more about API compatibility. Do we implement all the API that we currently rely on? I am not sure if this is the best approach to show or proof that. Open to ideas!


test/compatibility/dogstatsd_datagram_compatibility_test.rb
```diff
@@ -0,0 +1,119 @@
...
+      # until we receive a packet on the server end.
+      Thread.abort_on_exception = true
+      emitter_thread = Thread.new { loop { yield(client) } }
+
+      # Block until we read a packet, then kill the emitter thread.
+      data, _origin = @server.recvfrom(100)
+      emitter_thread.kill
+
+      # Now return the datagram
+      StatsD::Instrument::Datagram.new(data)
```
dylanahsmith: Looks like the tests are failing because it is comparing the Datagram objects, which don't implement `==`.  Why not compare the string packet itself?

wvanbergen: This depends on #191, which I extracted into a separate PR.

wvanbergen: Rebased

PR #193, opened 2019-09-30
Refactor assert_statsd_calls to flunk if an exception occurs inside the block

wvanbergen:
Ref #166 & #184.

This is a more straightforward way of dealing with exception that our previous approach. We tell people they should nest `assert_raises` inside the statsD assertion block.

This is a backwards incompatible change though - it may cause some tests to fail that previously were succeeding. However, the fix for those is easy: change the nesting order. The assertion message will tell you how to fix it.

``` ruby
# - This used to succeed in version < 2.3.3
# - This fails due to the lack of statsd increment in version >= 2.3.3
# - This will fail due to an unhandled exception after this PR is merged
  assert_statsd_increment('foo') do
    raise 'error'
  end
end

# - This used to succeed in version < 2.3.3
# - This also succeeds in version >= 2.3.3
# - This will fail due to an unhandled exception after this PR is merged
  assert_statsd_increment('foo') do
    StatsD.increment('foo')
    raise 'error'
  end
end

# - This behaves the same (succeeds) in all versions
assert_statsd_increment('foo') do
    StatsD.increment('foo')
    raise 'error'
  end
end

# - This behaves the same (fails) in all versions
assert_statsd_increment('foo') do
    raise 'error'
  end
end

```

### DIFF
--- a/test/assertions_test.rb
+++ b/test/assertions_test.rb
@@ -316,38 +316,55 @@ class AssertionsTest < Minitest::Test
     end
   end
 
-  def test_assertion_with_exceptions
+  def test_assertion_block_with_expected_exceptions
     assert_no_assertion_triggered do
+      @test_case.assert_statsd_increment('expected_happened') do
+        @test_case.assert_raises(RuntimeError) do
+          begin
+            raise "expected"
+          rescue
+            StatsD.increment('expected_happened')
+            raise
+          end
+        end
+      end
+    end
+
+    assertion = assert_assertion_triggered do
+      @test_case.assert_statsd_increment('counter') do
+        @test_case.assert_raises(RuntimeError) do
+          raise "expected"
+        end
+      end
+    end
+    assert_includes assertion.message, "No StatsD calls for metric counter of type c were made"
+  end
+
+  def test_assertion_block_with_unexpected_exceptions
+    assertion = assert_assertion_triggered do
+      @test_case.assert_statsd_increment('counter') do
+        StatsD.increment('counter')
+        raise "unexpected"
+      end
+    end
+    assert_includes assertion.message, "An exception occurred in the block provided to the StatsD assertion"
+
+    assertion = assert_assertion_triggered do
       @test_case.assert_raises(RuntimeError) do
         @test_case.assert_statsd_increment('counter') do
           StatsD.increment('counter')
-          raise "foo"
+          raise "unexpected"
         end
       end
     end
+    assert_includes assertion.message, "An exception occurred in the block provided to the StatsD assertion"
+  end
 
-    assert_no_assertion_triggered do
+  def test_assertion_block_with_other_assertion_failures
+    # If another assertion failure happens inside the block, that failrue should have priority
+    assert_assertion_triggered("other assertion failure") do
       @test_case.assert_statsd_increment('counter') do
-        @test_case.assert_raises(RuntimeError) do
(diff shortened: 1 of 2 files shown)


test/assertions_test.rb
```diff
@@ -316,40 +316,57 @@ def test_nested_assertions
...
 
-  def test_assertion_with_exceptions
+  def test_assertion_block_with_expected_exceptions
     assert_no_assertion_triggered do
-      @test_case.assert_raises(RuntimeError) do
-        @test_case.assert_statsd_increment('counter') do
-          StatsD.increment('counter')
-          raise "foo"
+      @test_case.assert_statsd_increment('expected_happened') do
+        @test_case.assert_raises(RuntimeError) do
```
dylanahsmith: Since the exception is rescued before getting to the code under test, does this half of the test actually test a code path that isn't tested elsewhere?

test/assertions_test.rb
```diff
@@ -316,40 +316,57 @@ def test_nested_assertions
...
+          rescue
+            StatsD.increment('expected_happened')
+            raise
+          end
         end
       end
     end
 
-    assert_no_assertion_triggered do
+    assertion = assert_assertion_triggered do
```
dylanahsmith: These helpers seem overly complicated.  Seems like this could just be `exc = assert_raises(Minitest::Assertion) do` and the `assert_no_assertion_triggered` method could just be omitted, since the exception it is catching would fail the test if not caught.

test/assertions_test.rb
wvanbergen: From a code coverage point of view, you could rewrite this as:
``` ruby
@test_case.assert_statsd_increment('expected_happened') do
  StatsD.increment('expected_happened')
end
```
For that pattern we already have tests for elsewhere, so we could :fire: this from a code coverage point of view.

However, I see this test more as documentation. The exception handling behavior of these assertion methods are not very intuitive, which explains why we keep tweaking these methods. So being a bit more explicit about how things will behave is a good thing.

dylanahsmith: If it is documentation, then I don't think this is the most discoverable place for it.  We are currently missing documentation on `assert_statsd_increment`, so we could actually document it and show this patterns as an example in that documentation.  Alternatively, we could provide the example in the failure message for when an unexpected exception occurs inside an `assert_statsd_increment` block.

I think the exception handling behaviour in this case is intuitive.  Without knowing anything about this gem, a ruby developer could tell that

```ruby
          begin
            raise "expected"
          rescue
            StatsD.increment('expected_happened')
            raise
          end
        end
```

would be equivalent to `StatsD.increment('expected_happened')`, unless `StatsD.increment('expected_happened')` raises another exception.  That seems like a convoluted way of testing that though, so also doesn't serve well as an example of how to test something.

wvanbergen: I like the example, because the `assert_statsd_increment` + `assert_raises` combination is an extremely common case in real application code (count the number of exceptions that occur), and where we exposed quite a lot of bugs since we shipped #166. 

The [README](https://github.com/Shopify/statsd-instrument#minitest) documents `assert_statsd_instrument` to some degree. We should really add YARD documentation to the assertion class.

wvanbergen:
Pushed a change to simplify the tests by getting rid of the helpers. Docs should be improved but I feel that's a bit outside the scope of this PR.

PR #194, opened 2019-09-30
Add Client#latency.

wvanbergen:
It measures the latency (execution time) of the provided block in milliseconds, and emits this value as a metric. The metric type by default is an `ms`, but can be overridden. For DogStatsD we default to a distribution metric.

We also use the latency method under the hood when `measure` or `distribution is called with a block, for backwards compatibility.

### DIFF
--- a/lib/statsd/instrument/client.rb
+++ b/lib/statsd/instrument/client.rb
@@ -76,7 +76,8 @@ class StatsD::Instrument::Client
   # @param sample_rate (see #increment)
   # @param tags (see #increment)
   # @return [void]
-  def measure(name, value = nil, sample_rate: nil, tags: nil)
+  def measure(name, value = nil, sample_rate: nil, tags: nil, &block)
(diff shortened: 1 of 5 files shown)


lib/statsd/instrument/client.rb
```diff
@@ -151,6 +153,30 @@ def histogram(name, value, sample_rate: nil, tags: nil)
...
+  # @param sample_rate (see #increment)
+  # @param tags (see #increment)
+  # @param [Symbol] metric_type The metric type to use. If not specified, we will
+  #   use the preferred metric type of the implementation. The default is `:ms`.
+  #   Generally, you should not have to set this.
+  # @yield The latency (execution time) of the block
+  # @return The return value of the proivded block will be passed through.
+  def latency(name, sample_rate: nil, tags: nil, metric_type: nil, &block)
+    start = Process.clock_gettime(Process::CLOCK_MONOTONIC)
+    block.call
```
casperisfine: use `yield` and remove `&block` from the signature. You'll save one `Proc` allocation.

`&block` is only worth it if you pass it to another method.

PR #196, opened 2019-10-01
Add support for Client#event and Client#service_check 

wvanbergen:
Those methods are only supported in DogStatsD, so will only be available when using the DogStatsD datagram builder. See https://docs.datadoghq.com/developers/dogstatsd/datagram_shell/ for the datagram format.

Questions:

- Note that neither of these support sample rates in their datagram formats. We could still apply sampling if we want to, but I think that's generally not what you'd expect. What do you think?
- Related, I don't expect these metrics to be emitted very often, and we can be a bit more strict w.r.t. input validation. Anything that we should check?
- Should we still use prefixes when one is defined on the client?

TODO in followups or in this PR: 

- ~Ensure the legacy API is compatible with the new one. (The old API is kind of awful, but we should do our best)~ A compatibility test has been added.
- Ensure the datagram parser can parse these datagrams.

### DIFF
--- a/lib/statsd/instrument/dogstatsd_datagram_builder.rb
+++ b/lib/statsd/instrument/dogstatsd_datagram_builder.rb
@@ -55,12 +55,12 @@ class StatsD::Instrument::DogStatsDDatagramBuilder < StatsD::Instrument::Datagra
   #
   # @see https://docs.datadoghq.com/developers/dogstatsd/datagram_shell/#service-checks
   def _sc(name, status, timestamp: nil, hostname: nil, tags: nil, message: nil)
-    status_number = SERVICE_CHECK_STATUS_VALUES.fetch(status)
+    status_number = status.is_a?(Integer) ? status : SERVICE_CHECK_STATUS_VALUES.fetch(status.to_sym)
     tags = normalize_tags(tags) + default_tags
 
     datagram = +"_sc|#{@prefix}#{normalize_name(name)}|#{status_number}"
-    datagram << "|d:#{timestamp.to_i}" if timestamp
     datagram << "|h:#{hostname}" if hostname
+    datagram << "|d:#{timestamp.to_i}" if timestamp
     datagram << "|##{tags.join(',')}" unless tags.empty?
     datagram << "|m:#{normalize_name(message)}" if message
     datagram
--- a/lib/statsd/instrument/dogstatsd_datagram_builder.rb
+++ b/lib/statsd/instrument/dogstatsd_datagram_builder.rb
@@ -27,7 +27,7 @@ class StatsD::Instrument::DogStatsDDatagramBuilder < StatsD::Instrument::Datagra
   def _e(title, text, timestamp: nil, hostname: nil, aggregation_key: nil, priority: nil,
     source_type_name: nil, alert_type: nil, tags: nil)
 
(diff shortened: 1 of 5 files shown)


lib/statsd/instrument/dogstatsd_datagram_builder.rb
```diff
@@ -8,4 +8,64 @@ class StatsD::Instrument::DogStatsDDatagramBuilder < StatsD::Instrument::Datagra
...
+    datagram = +"_sc|#{normalize_name(name)}|#{status_number}"
+    datagram << "|d:#{timestamp.to_i}" if timestamp
+    datagram << "|h:#{hostname}" if hostname
+    datagram << "|##{tags.join(',')}" unless tags.empty?
+    datagram << "|m:#{normalize_name(message)}" if message
+    datagram
+  end
+
+  SERVIC_CHECK_STATUS_VALUES = { ok: 0, warning: 1, critical: 2, unknown: 3 }.freeze
+  private_constant :SERVIC_CHECK_STATUS_VALUES
```
casperisfine: typo: `SERVIC_`

lib/statsd/instrument/dogstatsd_datagram_builder.rb
```diff
@@ -8,4 +8,64 @@ class StatsD::Instrument::DogStatsDDatagramBuilder < StatsD::Instrument::Datagra
...
+  #
+  # @see https://docs.datadoghq.com/developers/dogstatsd/datagram_shell/#service-checks
+  def _sc(name, status, timestamp: nil, hostname: nil, tags: nil, message: nil)
+    status_number = SERVICE_CHECK_STATUS_VALUES.fetch(status)
+    tags = normalize_tags(tags) + default_tags
+
+    datagram = +"_sc|#{normalize_name(name)}|#{status_number}"
+    datagram << "|d:#{timestamp.to_i}" if timestamp
+    datagram << "|h:#{hostname}" if hostname
+    datagram << "|##{tags.join(',')}" unless tags.empty?
```
dylanahsmith: I'm not sure if the order matters to datadog, but this was last in `StatsD::Instrument::Backends::UDPBackend::DogStatsDProtocol`, so probably should be here so that a DogStatsDDatagramCompatibilityTest test could be written for it.

wvanbergen: Interesting. This is indeed a difference. The DogstatsD documentation is quite explicit that `m` has to be the final parameter though:

> Add a message describing the current state of the service check. This field MUST be positioned last among the metadata fields. No default.

I have confirmed with the compatibility test that the behavior is actually different, but I am pretty sure this is a bug in the old implementation.

dylanahsmith: :ok:  we might not have actually used the tags with this

wvanbergen: #198 and #199 will fix some issues with the current implementation,, after which I can add a compatibility test.

wvanbergen:
After refactoring the current implementation of `StatsD.event` and `StatsD.service_check` to make sure they are actually according to spec, I was able to add a compatibility test that affirms the different implementations now behave the same.

This is ready for a final look.

PR #197, opened 2019-10-01
StatsD::Instrument::Environment improvements

wvanbergen:
- Make `StatsD::Instrument::Environment` a class rather than a singleton module, but make sure that `environment` and `default_backend` still work as singleton methods.
- Add `StatsD::Instrument::Environment#default_client` to instantiate a new Client based on  environment variables. 
- Respect the existing environment variables, but also add support for `STATSD_PREFIX` and `STATSD_DEFAULT_TAGS`. 
  The goal is that people should not configure StatsD in code, but completely using environment variables. If that doesn't work for your use case, we recommend that people instantiate their own client and not use the default client.

### DIFF
--- a/lib/statsd/instrument/environment.rb
+++ b/lib/statsd/instrument/environment.rb
@@ -4,31 +4,60 @@ require 'logger'
 
 # The environment module is used to detect, and initialize the environment in
 # which this library is active. It will use different default values based on the environment.
-module StatsD::Instrument::Environment
-  extend self
+class StatsD::Instrument::Environment
+  class << self
+    def from_env
+      @from_env ||= StatsD::Instrument::Environment.new(ENV)
+    end
 
-  # Instantiates a default backend for the current environment.
-  #
-  # @return [StatsD::Instrument::Backend]
-  # @see #environment
-  def default_backend
-    case environment
-    when 'production', 'staging'
-      StatsD::Instrument::Backends::UDPBackend.new(ENV['STATSD_ADDR'], ENV['STATSD_IMPLEMENTATION'])
-    when 'test'
-      StatsD::Instrument::Backends::NullBackend.new
-    else
-      StatsD::Instrument::Backends::LoggerBackend.new(StatsD.logger)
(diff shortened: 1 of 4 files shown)


lib/statsd/instrument/environment.rb
```diff
@@ -39,26 +68,65 @@ def datagram_builder_class
...
+  def statsd_implementation
+    env.fetch('STATSD_IMPLEMENTATION', 'statsd')
+  end
+
+  def statsd_sample_rate
+    env.fetch('STATSD_SAMPLE_RATE', 1.0).to_f
+  end
+
+  def statsd_prefix
+    env.fetch('STATSD_PREFIX', nil)
```
dylanahsmith: I thought you were planning on getting rid of the prefix?

wvanbergen: I don't think we can, because it is used all over the place. Most applications set a prefix to not interfere with each other.

wvanbergen:
The goal has always been that you can configure StatsD fully with environment variables, so that you don't have to use initializers, etc. In the past we used this to easily configure StatsD properly for all our heroku apps, now we can use this for cloud platform apps.

The Environment class/singleton does not hold configuration. Rather, it parses configuration from ENV. Because the logic is more complicated than simply reading environment variables, this has been extracted into a separate module.

- For the legacy client, some configuration values are set on the singleton (e.g. `StatsD.default_sample_rate`. This is done in the `Environment.setup`. Some other configuration is used to instantiate the backend that will be used as the singleton's backend (`StatsD.backend`)
- For the new `Client` implementation, the client instance holds all the configuration they need to operate. I would still like to be able to construct a client based on what is in the environment, and use that as the client that we use to delegate metric calls on the singleton to (e.g. `StatsD.increment`). 

I prefer using a class rather than a singleton for this because it easier to test (no need to stub values in `ENV`, etc). 

I agree that the implementation is a bit messy because it serves both the legacy client as well as the new client, and tries to be backwards compatible. Once we drop the legacy client, we can clean this class up.

PR #202, opened 2019-10-01
Respect global StatsD.prefix changes when using metaprogramming methods.

wvanbergen:
For the metaprogramming methods, we currently evaluate `StatsD.prefix` when the code is loading. If the prefix changes afterwards (for instance, in a Railtie or initializer), that new prefix will not be respected. This PR ensures that the new prefix will be respected.

I've also added some tests that demonstrate that you can override or clear the prefix when using the metaprogramming methods. When you do that, any changes to the global prefix will not have any affect on those metrics.

### DIFF
--- a/test/statsd_instrumentation_test.rb
+++ b/test/statsd_instrumentation_test.rb
@@ -337,6 +337,48 @@ class StatsDInstrumentationTest < Minitest::Test
     ActiveMerchant::Gateway.singleton_class.statsd_remove_count :sync, 'ActiveMerchant.Gateway.sync'
   end
 
+  def test_statsd_respects_global_prefix_changes
+    StatsD.prefix = 'Foo'
+    ActiveMerchant::Gateway.singleton_class.extend StatsD::Instrument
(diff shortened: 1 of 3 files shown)


test/statsd_instrumentation_test.rb
```diff
@@ -337,6 +337,48 @@ def test_statsd_count_with_tags
...
+    ActiveMerchant::Gateway.singleton_class.statsd_remove_count :sync, 'ActiveMerchant.Gateway.sync'
+  end
+
+  def test_statsd_macro_can_overwrite_prefix
+    StatsD.prefix = 'Foo'
+    ActiveMerchant::Gateway.singleton_class.extend StatsD::Instrument
+    ActiveMerchant::Gateway.singleton_class.statsd_measure :sync, 'ActiveMerchant.Gateway.sync', prefix: 'Bar'
+    StatsD.prefix = 'Quc'
+
+    statsd_calls = capture_statsd_calls { ActiveMerchant::Gateway.sync }
```
wvanbergen: I am not using `assert_statsd_measure` here because it automatically accounts for the prefix, which is very unhelpful in this case.

wvanbergen:
Released as 2.5.1

PR #203, opened 2019-10-02
Improve datagram compatibility test

wvanbergen:
We now run all the compatibility tests in the following modes:

- Normal (no prefix or default tags)
- With prefix
- With default tags

We now also have tests for `measure` and `distribution` in block execution time measuring mode.

There is also a reliability improvement by making sure that we always emit a packet by stubbing rand. That way, we can simply call the StatsD method once, rather than in a loop, and get rid of all the thread and draining logic.

### DIFF
--- a/lib/statsd/instrument/client.rb
+++ b/lib/statsd/instrument/client.rb
@@ -169,15 +169,17 @@ class StatsD::Instrument::Client
   # @return The return value of the proivded block will be passed through.
   def latency(name, sample_rate: nil, tags: nil, metric_type: nil)
     start = Process.clock_gettime(Process::CLOCK_MONOTONIC)
-    yield
-  ensure
-    stop = Process.clock_gettime(Process::CLOCK_MONOTONIC)
+    begin
+      yield
+    ensure
+      stop = Process.clock_gettime(Process::CLOCK_MONOTONIC)
 
-    sample_rate ||= @default_sample_rate
-    if sample?(sample_rate)
-      metric_type ||= datagram_builder.latency_metric_type
-      latency_in_ms = 1000.0 * (stop - start)
-      emit(datagram_builder.send(metric_type, name, latency_in_ms, sample_rate, tags))
+      sample_rate ||= @default_sample_rate
+      if sample?(sample_rate)
+        metric_type ||= datagram_builder.latency_metric_type
+        latency_in_ms = 1000.0 * (stop - start)
+        emit(datagram_builder.send(metric_type, name, latency_in_ms, sample_rate, tags))
+      end
     end
   end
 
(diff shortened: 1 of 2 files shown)


test/compatibility/dogstatsd_datagram_compatibility_test.rb
```diff
@@ -75,28 +108,32 @@ def with_legacy_client
...
-    rescue
-      IO::EAGAINWaitReadable
+      retried = false
+      loop do
+        @server.recvfrom_nonblock(100)
+        retried = false
+      end
+    rescue IO::WaitReadable
+      retried = true
+      retry unless retried
```
casperisfine: Hum, this is fishy. You set `retried = true` just above, so you'll never `retry` AFAICT.

Also when `IO::EAGAINWaitReadable` is raised you're suppose to wait with `IO.select`

test/compatibility/dogstatsd_datagram_compatibility_test.rb
```diff
@@ -17,56 +17,89 @@ def teardown
...
+      end
     end
 
     private
 
-    def with_legacy_client
+    MODES = [:normal, :with_prefix, :with_default_tags]
+
+    def assert_equal_datagrams(&block)
+      MODES.each do |mode|
```
casperisfine: So you edited your PR.

I think dynamically defining a test for each mode would be preferable:

```ruby
MODES.each do |mode|
  define_method("test_blah_in_#{mode}_mode") do
    # test
  end
end
```

This way if you break one of the mode it's much more clear.

wvanbergen:
I've simplified the implementation significantly, by stubbing `rand` (thanks Jean for the idea) so we are sure that every StatsD invocation will actually emit a datagram.

lib/statsd/instrument/client.rb
```diff
@@ -169,15 +169,17 @@ def histogram(name, value, sample_rate: nil, tags: nil)
...
-    if sample?(sample_rate)
-      metric_type ||= datagram_builder.latency_metric_type
-      latency_in_ms = 1000.0 * (stop - start)
-      emit(datagram_builder.send(metric_type, name, latency_in_ms, sample_rate, tags))
+      sample_rate ||= @default_sample_rate
+      if sample?(sample_rate)
+        metric_type ||= datagram_builder.latency_metric_type
+        latency_in_ms = 1000.0 * (stop - start)
+        emit(datagram_builder.send(metric_type, name, latency_in_ms, sample_rate, tags))
+      end
```
wvanbergen: This is a small but unrelated fix. While trying to get the compat tests to pass, I got a failed run because `start` was `nil` and `stop - start` raised an exception. I assume this means that `Process.clock_gettime` can fail sometimes? This change will make that that error more explicit if it happens again.

test/compatibility/dogstatsd_datagram_compatibility_test.rb
wvanbergen: I've tried this, but the code gets quite painful because we have to modify the StatsD singleton when we switch modes, and put it back in its original state.

Instead, I've made sure that if a datagram is not equal, the mode is included in the assertion method.

PR #206, opened 2019-10-02
Deprecate setting prefix on metric calls

wvanbergen:
- Deprecate `StatsD.measure(..., as_dict: true)`. AFAICS this keyword argument was mostly meant for internal use, but is part of a public method signature. You should use `StatsD.distribution` directly instead.
- Deprecate `StatsD.metric(..., prefix: "foo")`. You can simply incorporate the prefix in the metric name. If you want to override the global prefix, set `no_prefix` to `true` and incorporate your desired prefix in the name directly.

I've modified strict mode to blow up when these keyword arguments are used. Testing that on the Shopify codebase now.

### DIFF
--- a/lib/statsd/instrument/strict.rb
+++ b/lib/statsd/instrument/strict.rb
@@ -24,7 +24,7 @@ module StatsD
     #
     # This monkeypatch is not meant to be used in production.
     module Strict
-      def increment(key, value = 1, sample_rate: nil, tags: nil, prefix: StatsD.prefix, no_prefix: false)
+      def increment(key, value = 1, sample_rate: nil, tags: nil, no_prefix: false)
         raise ArgumentError, "StatsD.increment does not accept a block" if block_given?
         raise ArgumentError, "The value argument should be an integer, got #{value.inspect}" unless value.is_a?(Numeric)
         check_tags_and_sample_rate(sample_rate, tags)
@@ -32,7 +32,7 @@ module StatsD
         super
       end
 
-      def gauge(key, value, sample_rate: nil, tags: nil, prefix: StatsD.prefix, no_prefix: false)
+      def gauge(key, value, sample_rate: nil, tags: nil, no_prefix: false)
         raise ArgumentError, "StatsD.increment does not accept a block" if block_given?
         raise ArgumentError, "The value argument should be an integer, got #{value.inspect}" unless value.is_a?(Numeric)
         check_tags_and_sample_rate(sample_rate, tags)
@@ -40,7 +40,7 @@ module StatsD
         super
       end
 
-      def histogram(key, value, sample_rate: nil, tags: nil, prefix: StatsD.prefix, no_prefix: false)
+      def histogram(key, value, sample_rate: nil, tags: nil, no_prefix: false)
         raise ArgumentError, "StatsD.increment does not accept a block" if block_given?
         raise ArgumentError, "The value argument should be an integer, got #{value.inspect}" unless value.is_a?(Numeric)
         check_tags_and_sample_rate(sample_rate, tags)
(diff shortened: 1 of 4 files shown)


casperisfine:
> Ultimately different prefix should be different clients.

That or go `redis-namespace` style.

wvanbergen:
> Ultimately different prefix should be different clients. 

That's the plan with the new client:

``` ruby
prefixed_client = StatsD.client.with_prefix('foo')
prefixed_client.increment('bar')
```

lib/statsd/instrument/strict.rb
```diff
@@ -111,31 +107,113 @@ def collect_metric(type, name, value, sample_rate:, tags: nil, prefix:, metadata
...
 
     module StrictMetaprogramming
-      def statsd_measure(method, name, sample_rate: nil, tags: nil,
-        prefix: StatsD.prefix, no_prefix: false, as_dist: false)
-
+      def statsd_measure(method, name, sample_rate: nil, tags: nil, no_prefix: false)
         check_method_and_metric_name(method, name)
-        super
+
+        # Unfortunately, we have to inline the new method implementation ebcause we have to fix the
```
dylanahsmith:
```suggestion
        # Unfortunately, we have to inline the new method implementation because we have to fix the
```

this typo was also copied to following methods

lib/statsd/instrument/strict.rb
```diff
@@ -111,31 +107,113 @@ def collect_metric(type, name, value, sample_rate:, tags: nil, prefix:, metadata
...
+        # Stats.measure call to not use the `as_dist` and `prefix` arguments.
```
dylanahsmith: I think you mean

```suggestion
        # statsd_measure call to not use the `as_dist` and `prefix` arguments.
```

Although, couldn't we change the non-strict version of `statsd_measure` to use a keyword argument splat to pass through deprecated parameters?  If it is passing those parameters to a strict singleton method, then I don't think we would even have to override these method wrappers for strictness, unless we wanted to make sure that the arguments are checked at initialization time.

dylanahsmith: For example, we could do this to avoid passing `prefix` or `as_dist` arguments while still handling deprecated positional arguments in the same way.

```diff
diff --git a/lib/statsd/instrument.rb b/lib/statsd/instrument.rb
index d32b34d..14412eb 100644
--- a/lib/statsd/instrument.rb
+++ b/lib/statsd/instrument.rb
@@ -90,15 +90,14 @@ module StatsD
     #    callable to dynamically generate a metric name
     # @param metric_options (see StatsD#measure)
     # @return [void]
-    def statsd_measure(method, name, deprecated_sample_rate_arg = nil, deprecated_tags_arg = nil, as_dist: false,
-      sample_rate: deprecated_sample_rate_arg, tags: deprecated_tags_arg, prefix: nil, no_prefix: false)
+    def statsd_measure(method, name, deprecated_sample_rate_arg = nil, deprecated_tags_arg = nil,
+      sample_rate: deprecated_sample_rate_arg, tags: deprecated_tags_arg, **statsd_kwargs)
 
       add_to_method(method, name, :measure) do
         define_method(method) do |*args, &block|
           key = StatsD::Instrument.generate_metric_name(name, self, *args)
-          prefix ||= StatsD.prefix
           StatsD.measure(
-            key, sample_rate: sample_rate, tags: tags, prefix: prefix, no_prefix: no_prefix, as_dist: as_dist
+            key, sample_rate: sample_rate, tags: tags, **statsd_kwargs
           ) do
             super(*args, &block)
           end
```

wvanbergen: It would be a minor performance hit right, because we would use a keyword splat? It does clean up the code quite nicely.

dylanahsmith: Oh, it looks like you are right that it would be a minor performance hit.   I expected that capturing the keyword arguments would allocate a hash, but that happens at initialization time.  But it looks like applying the keyword arguments hash actually results in a couple of allocations.

That seems like something that could be optimized in ruby itself.  It is unfortunate that we have a trade-off between code and runtime complexity with the current state of ruby.

wvanbergen: Yup, for the time being this strict mode file is pretty awful, but at least it's contained to this file. I'd prefer the 💩 not to spread outside of this file, even if that means to take on some extra duplication.

PR #207, opened 2019-10-02
Add StatsD.client singleton

wvanbergen:
Expose a singleton client (by default, instantiated using environment variables), so in the future we can delegate the singleton methods like `StatsD.increment` to it.

This will allow us to more easily test compatibility with the new client. We may also need #171 for that.

### DIFF
--- a/lib/statsd/instrument.rb
+++ b/lib/statsd/instrument.rb
@@ -329,7 +329,7 @@ module StatsD
   end
 
   attr_accessor :logger, :default_sample_rate, :prefix
-  attr_writer :backend
+  attr_writer :backend, :client
   attr_reader :default_tags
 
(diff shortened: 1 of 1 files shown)


lib/statsd/instrument.rb
```diff
@@ -340,6 +340,11 @@ def backend
     @backend ||= StatsD::Instrument::Environment.default_backend
   end
 
+  def client
+    require 'statsd/instrument/client'
```
casperisfine: Might want to but that require in a `begin/end`. Otherwise every access to `StatsD.client` will have to do a lookup in `$LOADED_FEATURES`, which isn't free.

wvanbergen: Fixed!

PR #208, opened 2019-10-02
Delegate the decision whether to sample a metric to the sink.

wvanbergen:
Delegate the decision whether to sample a metric to the sink. This allows us to sink to determine whether it wants to see the packet.

- `NullSink`: never
- `LogSink`: always
- `CaptureSink`: always
- `UDPSink`: implement actual sampling using `rand()`

This will allow us to use `capture` to capture packets like the old client.

### DIFF
--- a/test/client_test.rb
+++ b/test/client_test.rb
@@ -109,10 +109,9 @@ class ClientTest < Minitest::Test
 
   def test_latency_calls_block_even_when_not_sending_a_sample
     called = false
-    datagrams = @client.capture do
+    @client.capture do
       @client.latency('foo', sample_rate: 0) { called = true }
     end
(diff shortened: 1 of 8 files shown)


test/client_test.rb
```diff
@@ -109,23 +109,23 @@ def test_latency_on_dogstatsd_prefers_distribution_metric_type
 
   def test_latency_calls_block_even_when_not_sending_a_sample
     called = false
-    datagrams = @client.capture do
+    @client.capture do
       @client.latency('foo', sample_rate: 0) { called = true }
     end
-    assert_predicate datagrams, :empty?
```
wvanbergen: Note: because we use `capture`, the datagram will now always be captured no matter what the sample rate is.

PR #209, opened 2019-10-02
New client: add support for `no_prefix: true` when emitting metrics.

wvanbergen:
I am not a big fan of the `no_prefix` keyword argument. With the new client, it's easy to instantiate another client based on the current client with a different prefix, and use this separate client to emit your metrics.

However, I don’t think we can deprecate it with the new client. Specifically some libraries depend on this, and they may have to support `statsd-instrument` v2 and v3 simultaneously during the upgrade period. So I think we'll have to support this option.

### DIFF
--- a/lib/statsd/instrument/client.rb
+++ b/lib/statsd/instrument/client.rb
@@ -30,6 +30,8 @@ class StatsD::Instrument::Client
     @prefix = prefix
     @default_tags = default_tags
     @default_sample_rate = default_sample_rate
+
+    @datagram_builder = { false => nil, true => nil }
   end
 
   # @!group Metric Methods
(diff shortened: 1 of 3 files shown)


lib/statsd/instrument/client.rb
```diff
@@ -282,8 +295,11 @@ def capture(&block)
 
   protected
 
-  def datagram_builder
-    @datagram_builder ||= @datagram_builder_class.new(prefix: prefix, default_tags: default_tags)
+  def datagram_builder(no_prefix:)
```
wvanbergen: Adding the no-prefix option here limits the scope of adding support for this option: the datagram builder class doens't have to be aware of its existence. It's not super super elegant though :/

PR #210, opened 2019-10-02
Rubocop: check for as_dist: true in `StatsD.measure` and `statsd_measure`

wvanbergen:
Using `as_dist: true` is deprecated as argument to `StatsD.measure` (see #206).

This adds a Rubocop rule that checks for this usage, and tells people to change it to `StatsD.distribution`.  It does the same for the `statsd_measure` metaprogramming method.


test/statsd_instrumentation_test.rb
```diff
@@ -248,19 +248,6 @@ def test_statsd_measure_with_sample_rate
...
-    ActiveMerchant::UniqueGateway.statsd_measure :ssl_post, 'ActiveMerchant.Gateway.ssl_post', as_dist: true
-
-    assert_statsd_distribution('ActiveMerchant.Gateway.ssl_post') do
-      ActiveMerchant::UniqueGateway.new.purchase(true)
-    end
-  ensure
-    unless StatsD::Instrument.strict_mode_enabled?
-      ActiveMerchant::UniqueGateway.statsd_remove_measure :ssl_post, 'ActiveMerchant.Gateway.ssl_post'
-    end
-  end
```
wvanbergen: This is now tested in deprecations_test.rb

test/statsd_instrumentation_test.rb
```diff
@@ -354,24 +341,6 @@ def test_statsd_respects_global_prefix_changes
...
-
-    statsd_calls = capture_statsd_calls { ActiveMerchant::Gateway.sync }
-    assert_equal 1, statsd_calls.length
-    assert_equal "Bar.ActiveMerchant.Gateway.sync", statsd_calls.first.name
-  ensure
-    unless StatsD::Instrument.strict_mode_enabled?
-      StatsD.prefix = nil
-      ActiveMerchant::Gateway.singleton_class.statsd_remove_measure :sync, 'ActiveMerchant.Gateway.sync'
-    end
-  end
```
wvanbergen: This is now tested in deprecations_test.rb

test/statsd_test.rb
```diff
@@ -239,14 +219,6 @@ def test_name_prefix
...
     m = capture_statsd_call { StatsD.increment('counter', no_prefix: true) }
     assert_equal 'counter', m.name
-
-    unless StatsD::Instrument.strict_mode_enabled?
-      m = capture_statsd_call { StatsD.increment('counter', prefix: "foobar") }
-      assert_equal 'foobar.counter', m.name
-
-      m = capture_statsd_call { StatsD.increment('counter', prefix: "foobar", no_prefix: true) }
-      assert_equal 'counter', m.name
-    end
```
wvanbergen: This is split off in a separate test and moved to deprecations_test.rb

test/statsd_test.rb
```diff
@@ -44,25 +38,11 @@ def test_statsd_measure_with_benchmarked_block_duration
...
     return_value = StatsD.measure('values.foobar') { 'sarah' }
     assert_equal 'sarah', return_value
   end
 
-  def test_statsd_measure_as_distribution_returns_return_value_of_block_even_if_nil
-    skip("StatsD.measure(..., as_dist: true) is deprecated") if StatsD::Instrument.strict_mode_enabled?
-    return_value = StatsD.measure('values.foobar', as_dist: true) { nil }
-    assert_nil return_value
-  end
-
```
wvanbergen: This, and the tests above, now live in deprecations_test.rb

lib/statsd/instrument/rubocop/metric_value_keyword_argument.rb
```diff
@@ -1,35 +1,35 @@
...
-
-              check_keyword_arguments_for_value_entry(node, last_argument) if last_argument&.type == :hash
+          if metric_method?(node)
+            last_argument = if node.arguments.last&.type == :block_pass
+              node.arguments[node.arguments.length - 2]
+            else
+              node.arguments[node.arguments.length - 1]
             end
+
+            check_keyword_arguments_for_value_entry(node, last_argument) if last_argument&.type == :hash
```
dylanahsmith: Why not use the new `keyword_arguments` you added to the `RuboCop::Cop::StatsD` module.

```suggestion
            keyword_args = keyword_arguments(node)
            check_keyword_arguments_for_value_entry(node, keyword_args) if keyword_args
```

lib/statsd/instrument/rubocop/metric_prefix_argument.rb
```diff
@@ -0,0 +1,45 @@
...
+                pair.child_nodes[0].value == :prefix
+            end
+            add_offense(node) if prefix
+          end
+
+          if metaprogramming_method?(node) && (hash = keyword_arguments(node))
+            prefix = hash.child_nodes.detect do |pair|
+              pair.child_nodes[0]&.type == :sym &&
+                pair.child_nodes[0].value == :prefix
+            end
```
dylanahsmith: This pattern is repeated 4 times in this PR.  Perhaps there should be a helper method for getting a specific keyword argument from the keyword arguments node.

lib/statsd/instrument.rb
```diff
@@ -97,7 +97,7 @@ def statsd_measure(method, name, deprecated_sample_rate_arg = nil, deprecated_ta
         define_method(method) do |*args, &block|
           key = StatsD::Instrument.generate_metric_name(name, self, *args)
           prefix ||= StatsD.prefix
-          StatsD.measure(
+          StatsD.measure( # rubocop:disable StatsD/MeasureAsDistArgument, StatsD/MetricPrefixArgument
```
casperisfine: Not sure it makes sense to enable those cops on statsd-instrument itself.

lib/statsd/instrument/rubocop/metric_value_keyword_argument.rb
wvanbergen: Fixed

lib/statsd/instrument/rubocop/metric_prefix_argument.rb
wvanbergen: Fixed

lib/statsd/instrument.rb
wvanbergen: I like it, because we make sure the rules actually work this way.

This particular exception can :fire: when we do the next major release.

wvanbergen:
Added some tests for the new rules, and refactored some stuff into helper methods.

PR #214, opened 2019-10-07
Support the new client when capturing datagrams from the singleton client

wvanbergen:
- Set up datagram capturing based on whether the singleton is the legacy singleton, or an instance of the new client.
- Update `assert_statsd_*` methods to work with both clients.
- Run the assertions tests both against the new and the legacy clients.
- Compatibility fixes between `Metric` & `Datagram` and how they interface with `MetricExpectation`.

The big difference is that the old capture method will return a bunch of `StatsD::Instrument::Metric` instances, while the new client will give a bunch of `StatsD::Instrument::Datagram` instances. These objects are not 100% compatible. The behave closely enough to pass the tests that we have for them, but I'll see if I can find a can run the Shopify CI suite to see if we depend on implementation details.


rafaelfranca:
LGTM. Let me know when it is out of draft so I can approve.

wvanbergen:
This is ready for review now. We might need some more compatibility fixes but the latest changes have addressed a vast majority of them.

PR #220, opened 2019-10-09
Bump to version 2.7.0

wvanbergen:
Add changelog entries describing the changes and new deprecations.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,76 @@ section below.
 
 _Nothing yet!_
 
+## Version 2.7.0
+
+This release allows you to switch the StatsD singleton to use the new, more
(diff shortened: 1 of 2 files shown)


CHANGELOG.md
```diff
@@ -8,6 +8,68 @@ section below.
 
 _Nothing yet!_
 
+## Version 2.6.1
```
dylanahsmith: Shouldn't this be 2.7.0 according to semantic versioning?  Since it adds to the API.

rafaelfranca: Yes.

wvanbergen: Updated. Also added the new rubocop rule.

PR #221, opened 2019-10-09
Add Rubocop rule that finds uses of StatsD singleton configuration methods

wvanbergen:
The following methods are deprecated because they configure the singleton, which we are moving away from:

- `StatsD.prefix`
- `StatsD.backend`
- `StatsD.default_sample_rate`
- `StatsD.default_tags`

This Rubocop rule finds uses of these methods and tells you how to fix.

### DIFF
--- a/lib/statsd/instrument/rubocop/singleton_configuration.rb
+++ b/lib/statsd/instrument/rubocop/singleton_configuration.rb
@@ -24,7 +24,7 @@ module RuboCop
       #   and assign it to `StatsD.singleton_client`. The client constructor accepts many of the
       #   same options.
       # - If you have to, you can call the old methods on `StatsD.legacy_singleton_client`. Note
-      #   that thios option will go away in the next major version.
+      #   that this option will go away in the next major version.
       class SingletonConfiguration < Cop
         include RuboCop::Cop::StatsD
 
--- a/lib/statsd/instrument/rubocop/singleton_configuration.rb
+++ b/lib/statsd/instrument/rubocop/singleton_configuration.rb
@@ -1,0 +1,53 @@
+# frozen-string-literal: true
+
+require_relative '../rubocop' unless defined?(RuboCop::Cop::StatsD)
+
+module RuboCop
+  module Cop
+    module StatsD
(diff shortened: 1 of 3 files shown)


lib/statsd/instrument/rubocop/singleton_configuration.rb
```diff
@@ -0,0 +1,53 @@
...
+
+          - The best option is to configure the library using environment variables, like
+            `STATSD_ADDR`, `STATSD_IMPLEMENTATION`, `STATSD_PREFIX`, and `STATSD_DEFAULT_TAGS`.
+            Metric methods called on the StatsD singleton (e.g. `StatsD.increment`) will by default
+            be delegated to a client that is configured using these environment variables.
+          - Alternatively, you can instantiate your own client using `StatsD::Instrument::Client.new`,
+            and assign it to `StatsD.singleton_client`. The client constructor accepts many of the
+            same options.
+          - If you have to, you can call the old methods on `StatsD.legacy_singleton_client`. Note
+            that thios option will go away in the next major version.
```
dylanahsmith:
```suggestion
            that this option will go away in the next major version.
```

lib/statsd/instrument/rubocop/singleton_configuration.rb
```diff
@@ -0,0 +1,53 @@
...
+      #
+      # - The best option is to configure the library using environment variables, like
+      #   `STATSD_ADDR`, `STATSD_IMPLEMENTATION`, `STATSD_PREFIX`, and `STATSD_DEFAULT_TAGS`.
+      #   Metric methods called on the StatsD singleton (e.g. `StatsD.increment`) will by default
+      #   be delegated to a client that is configured using these environment variables.
+      # - Alternatively, you can instantiate your own client using `StatsD::Instrument::Client.new`,
+      #   and assign it to `StatsD.singleton_client`. The client constructor accepts many of the
+      #   same options.
+      # - If you have to, you can call the old methods on `StatsD.legacy_singleton_client`. Note
+      #   that thios option will go away in the next major version.
```
dylanahsmith:
```suggestion
      #   that this option will go away in the next major version.
```

PR #228, opened 2019-10-17
Allow specifying an implementation when instantiating a Client

wvanbergen:
An `implementation` is easier to specify than a `datagram_builder_class`, and more closely aligned with how you would configure a client using environment variables.

We will pick the right datagram_builder_class based on the implementation, but you can still specify a datagram_builder_class yourself if you prefer.

### DIFF
--- a/lib/statsd/instrument/client.rb
+++ b/lib/statsd/instrument/client.rb
@@ -22,7 +22,8 @@ class StatsD::Instrument::Client
     prefix: nil,
     default_sample_rate: 1,
     default_tags: nil,
-    datagram_builder_class: StatsD::Instrument::StatsDDatagramBuilder
+    implementation: 'datadog',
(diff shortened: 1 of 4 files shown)


lib/statsd/instrument/client.rb
```diff
@@ -22,10 +22,11 @@ def initialize(
     prefix: nil,
     default_sample_rate: 1,
     default_tags: nil,
-    datagram_builder_class: StatsD::Instrument::StatsDDatagramBuilder
+    implementation: 'datadog',
+    datagram_builder_class: nil
```
dylanahsmith: You could handle the default here instead of in the method implementation

```suggestion
    datagram_builder_class: datagram_builder_class_from_implementation(implementation)
```

wvanbergen: :+1:

PR #230, opened 2019-10-18
Refactor constructing a client based on environment variables

wvanbergen:
Right now, you can instantiate a client based on the environment, or you can instantiate a client from scratch. What is not easy to to instantiate a client that uses most characteristics from the environment, but overrides some settings.

To do this you can either do:
- `Client.new`, and set all the keyword argument that you don't want to override with the values from the Environment object. This is painful, and if we ever add new keyword arguments that can be set based on the environment,  people need to update their code.
- `Environment.current.default_client.clone_with_options(...)`. This kind of works the way we want it to, but is not very discoverable and verbose.

The specific use case I have in mind is that for many apps, we instantiate the client from environment, but want to set a prefix manually based on the name of the application.

To make this easier, I have updated the default values for `Client#initialize` to not be hardcoded, but coming from the environment instead.

I am not quite sure if this is the right way of doing this, so feel free to disagree with this approach

### DIFF
--- a/lib/statsd/instrument.rb
+++ b/lib/statsd/instrument.rb
@@ -394,7 +394,7 @@ module StatsD
   end
 
   def client
-    @client ||= StatsD::Instrument::Environment.current.default_client
+    @client ||= StatsD::Instrument::Client.from_env
   end
 
   # Singleton methods will be delegated to the singleton client.
--- a/lib/statsd/instrument.rb
+++ b/lib/statsd/instrument.rb
@@ -390,11 +390,11 @@ module StatsD
   end
 
   def singleton_client
-    @singleton_client ||= StatsD::Instrument::Environment.from_env.client
+    @singleton_client ||= StatsD::Instrument::Environment.current.client
   end
 
   def client
-    @client ||= StatsD::Instrument::Environment.from_env.default_client
+    @client ||= StatsD::Instrument::Environment.current.default_client
   end
 
   # Singleton methods will be delegated to the singleton client.
--- a/lib/statsd/instrument.rb
+++ b/lib/statsd/instrument.rb
@@ -381,7 +381,7 @@ module StatsD
   end
 
   attr_accessor :logger
-  attr_writer :client, :singleton_client
+  attr_writer :singleton_client
 
   extend Forwardable
 
@@ -391,10 +391,6 @@ module StatsD
 
   def singleton_client
     @singleton_client ||= StatsD::Instrument::Environment.current.client
-  end
-
-  def client
-    @client ||= StatsD::Instrument::Client.from_env
   end
 
   # Singleton methods will be delegated to the singleton client.
--- a/lib/statsd/instrument/client.rb
+++ b/lib/statsd/instrument/client.rb
@@ -15,15 +15,48 @@ require 'statsd/instrument/log_sink'
 #   next major release of this library. While this class may already be functional,
 #   we provide no guarantees about the API and the behavior may change.
 class StatsD::Instrument::Client
+  class << self
+    def from_env(
+      env = StatsD::Instrument::Environment.current,
+      prefix: env.statsd_prefix,
+      default_sample_rate: env.statsd_sample_rate,
+      default_tags: env.statsd_default_tags,
+      implementation: env.statsd_implementation,
+      sink: env.default_sink_for_environment,
+      datagram_builder_class: datagram_builder_class_for_implementation(implementation)
+    )
+      new(
+        prefix: prefix,
+        default_sample_rate: default_sample_rate,
+        default_tags: default_tags,
+        implementation: implementation,
+        sink: sink,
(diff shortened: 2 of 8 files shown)


wvanbergen:
Alternatively, we could leave `new` as is with hardcoded defaults, but add an `from_env` factory method that works like this.

Edouard-chin:
> Alternatively, we could leave new as is with hardcoded defaults, but add an from_env factory method that works like this.

This seems like a better approach to me. I know that most of the time we build a client and need most of the characteristics from the env but for the rare occasions where that's not the case then we'd have the opposite issue with this patch.

Edouard-chin:
Also, I think we should deprecate usage of `clone_with_options` and get rid of it in the next major release. As you said its not very easy to discover and its purpose is gonna be useless if we decide to implement the `from_env` factory

wvanbergen:
It serves a different use case. It allows you to instantiate new clients that are the same as a different client, except for some options.

E.g., it allows you to instantiate a separate client for a specific purpose. E.g. background processing, for which we always set a `BackGroundQueue` prefix and the job class name as tag. Other than that, you want it to behave the same as the main client.

The idea would be to do something like this: `BackgroundQueue.statsd = StatsD.singleton_client.clone_with_options(prefix: ...)`

This is one of the main advantages of the new client not being a singleton anymore.

`clone_with_options` might not be the best name though?

Edouard-chin:
Ah yes ok make sense. I still think having `clone_with_options` is redundant if we have `from_env`.

What I propose is this:

```ruby
class Client
  def self.copy_from(from:, **options)
    case from
    when 'env'
      build(prefix: StatsD::Instrument::Environment.current.statsd_prefix, ...)
    when Client
      build(prefix: options[:prefix] || from.prefix, ...)
    end
  end

  def self.build(options)
    new(**options)
  end
end
```

wvanbergen:
I pushed a bunch of updates:

- `Client.new` is unchanged
- `Client.from_env` now allows you to construct a client based on the environment, and overwrite what you want to overwrite.
- A bunch of Client-logic that was previously in the Environment class has been removed, because we can the Client class directly now.
- I removed `StatsD.client`, which always had a new style client assigned to it. You should use `StatsD.singleton_client` (which can be a LegacyClient), or instantiate a client yourself and assign it to your own variable of choice.

lib/statsd/instrument/environment.rb
```diff
@@ -4,8 +4,8 @@
 # which this library is active. It will use different default values based on the environment.
 class StatsD::Instrument::Environment
   class << self
-    def from_env
-      @from_env ||= StatsD::Instrument::Environment.new(ENV)
+    def current
```
casperisfine: `current` is a bit weird of a name IMO. As it suggest it is likely to change like `User.current` would.

IMO something like `default` or `main` would make more sense.

lib/statsd/instrument.rb
```diff
@@ -390,11 +390,7 @@ def legacy_singleton_client
   end
 
   def singleton_client
-    @singleton_client ||= StatsD::Instrument::Environment.from_env.client
-  end
-
-  def client
```
Edouard-chin: Shouldn't we deprecated it first? Or a Changelog entry is enough you think ?

lib/statsd/instrument/client.rb
```diff
@@ -15,15 +15,48 @@
 #   next major release of this library. While this class may already be functional,
 #   we provide no guarantees about the API and the behavior may change.
 class StatsD::Instrument::Client
+  class << self
+    def from_env(
```
casperisfine: Lightly held opinion: In the vast majority of cases you want to use the ENV values as default. e.g. you are much more likely to simply want to have a different `prefix` and / or `default_tags`, than to ignore `$STATSD_ADDR`.

As such I think `.new` should use ENV default, and that another method (no idea of a name just like that) should allow to instantiate a new client without inheriting from `ENV`.

lib/statsd/instrument.rb
wvanbergen: I think we can get away with a CHANGELOG entry. Unless people went out of their way, they should not have been using this method.

lib/statsd/instrument/environment.rb
wvanbergen: I named it `current` because it is the "current environment". I don't like `default`, because that implies it's something hardcoded, which is not the case. Not too sure about `main`

lib/statsd/instrument/client.rb
wvanbergen: This is whatI had first, but changed it based on @Edouard-chin's comments. 

What I don't like about using the Environment object as default values for `initialize` is that it makes the object effective always dependent on another object. That might be dependency injection purism, but in many other instances I have been bitten by this because it increases the need for brittle mocking when writing unit tests.

PR #234, opened 2019-10-23
Add support for parsing service check and event datagrams.

wvanbergen:
We currently do not support parsing the custom datagrams that DogStatsD support, i.e. service checks and events. Note that we are perfectly able to send them. But if we capture them, we cannot parse them. This is mostly a problem for test suites that want to run assertions on properties of the datagrams.

This PR implements parsing the datagrams by subclassing Datagram, add regular expressions for those datagram types, and use `Regexp.union` to construct a parser that supports all of the datagrams.

Note: we only support parsing datagrams as generated by this library, not necessarily every datagram out there. Given the use case, that is fine I think.

### DIFF
--- a/lib/statsd/instrument/dogstatsd_datagram.rb
+++ b/lib/statsd/instrument/dogstatsd_datagram.rb
@@ -1,0 +1,88 @@
+# frozen_string_literal: true
+
+# The Datagram class parses and inspects a StatsD datagrams
+#
+# @note This class is part of the new Client implementation that is intended
+#   to become the new default in the next major release of this library.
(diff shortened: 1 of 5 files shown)


lib/statsd/instrument/dogstatsd_datagram.rb
```diff
@@ -0,0 +1,88 @@
...
+    else super
+    end
+  end
+
+  def hostname
+    parsed_datagram[:hostname]
+  end
+
+  def timestamp
+    Time.at(Integer(parsed_datagram[:timestamp]))
```
casperisfine: maybe add a `.utc`?

PR #235, opened 2019-10-23
Make assert_no_statsd_calls more flexible

sambostock:
Currently, `assert_no_statsd_calls` has two modes:
- match by exact single metric name
    ```ruby
    assert_no_statsd_calls('my.metric')
    ```
- match all metrics
    ```ruby
    assert_no_statsd_calls
    ```

This **adds** flexibility in the following ways:
- **Variadic arguments, to support exact matches against multiple names**
    ```ruby
    assert_no_statsd_calls('my.metric', 'your.metric')
    ```
    This is useful for reducing verbosity when multiple specific metrics need to be detected
<details>
<summary><strike>Support for matching by Regexp pattern</strike></summary>

```ruby
assert_no_statsd_calls(/\.metric/)
 ```
This is useful in cases where the number of metrics to watch out for is impractical, but detecting all metrics would be too restrictive
_e.g. any metric with a certain prefix_

</details>

<details>
<summary><strike>Any combination thereof</strike></summary>

```ruby
assert_no_statsd_calls('my.metric', /your/)
```

</details>

### DIFF
--- a/lib/statsd/instrument/assertions.rb
+++ b/lib/statsd/instrument/assertions.rb
@@ -44,21 +44,21 @@ module StatsD::Instrument::Assertions
 
   # Asserts no metric occurred during the execution of the provided block.
   #
-  # @param [String] metric_name (default: nil) The metric name that is not allowed
-  #   to happen inside the block. If this is set to `nil`, the assertion will fail
-  #   if any metric occurs.
+  # @param [Array<String>] metric_names (default: []) The metric names that are not
+  #   allowed to happen inside the block. If this is set to `[]`, the assertion
+  #   will fail if any metric occurs.
   # @yield A block in which the specified metric should not occur. This block
   #   should not raise any exceptions.
   # @return [void]
   # @raise [Minitest::Assertion] If an exception occurs, or if any metric (with the
-  #   provided name, or any), occurred during the execution of the provided block.
-  def assert_no_statsd_calls(metric_name = nil, datagrams: nil, client: nil, &block)
+  #   provided names, or any), occurred during the execution of the provided block.
+  def assert_no_statsd_calls(*metric_names, datagrams: nil, client: nil, &block)
     if datagrams.nil?
       raise LocalJumpError, "assert_no_statsd_calls requires a block" unless block_given?
       datagrams = capture_statsd_datagrams_with_exception_handling(client: client, &block)
     end
 
-    datagrams.select! { |m| m.name == metric_name } if metric_name
+    datagrams.select! { |metric| metric_names.include?(metric.name) } unless metric_names.empty?
     assert(datagrams.empty?, "No StatsD calls for metric #{datagrams.map(&:name).join(', ')} expected.")
   end
 
--- a/test/assertions_test.rb
+++ b/test/assertions_test.rb
@@ -33,6 +33,31 @@ class AssertionsTest < Minitest::Test
     end
     assert_equal assertion.message, "No StatsD calls for metric counter expected."
 
+    @test_case.assert_no_statsd_calls('counter1', 'counter2') do
+      # noop
+    end
+
+    @test_case.assert_no_statsd_calls('counter1', 'counter2') do
+      StatsD.increment('counter')
+    end
+
+    assertion = assert_raises(Minitest::Assertion) do
+      @test_case.assert_no_statsd_calls('counter1', 'counter2') do
+        StatsD.increment('counter0')
+        StatsD.increment('counter1')
+        StatsD.increment('counter2')
+        StatsD.increment('counter3')
+      end
+    end
+    assert_equal assertion.message, "No StatsD calls for metric counter1, counter2 expected."
+
+    assertion = assert_raises(Minitest::Assertion) do
+      @test_case.assert_no_statsd_calls('counter0', 'counter1', 'counter2') do
+        StatsD.increment('counter1')
+      end
+    end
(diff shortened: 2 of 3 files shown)


wvanbergen:
I am not a big fan of the regular expression option. When in practice do you need that?

lib/statsd/instrument/assertions.rb
```diff
@@ -44,21 +44,33 @@ module StatsD::Instrument::Assertions
...
-    datagrams.select! { |m| m.name == metric_name } if metric_name
+    unless metric_names_or_patterns.empty?
+      datagrams.select! do |metric|
+        metric_names_or_patterns.any? do |metric_name_or_pattern|
+          case metric_name_or_pattern
+          when Regexp
+            metric_name_or_pattern =~ metric.name
+          when String
+            metric_name_or_pattern == metric.name
+          end
```
sambostock: I originally implemented this as
```ruby
metric_name_or_pattern === metric.name
```
but Rubocop complained.

Technically, this implementation means that if someone was previously using something other than a `String`, it would no longer work, but I'm not sure how that would have been working in the first place, or if it should be supported. I could always switch to something like

```ruby
case metric_name_or_pattern
when Regexp
  metric_name_or_pattern =~ metric.name
else
  metric_name_or_pattern == metric.name
end
```
or
```ruby
if metric_name_or_pattern.is_a?(Regexp)
  metric_name_or_pattern =~ metric.name
else
  metric_name_or_pattern == metric.name
end
```

sambostock:
@wvanbergen I have a job which forwards some statistics we get from Twilio about our queues. I would like to use this change to do something like this:

```ruby
datagrams = capture_stats_d_calls do
  forward_statistics_from_twilio_to_datadog
end

assert_statsd_distribution('twilio.statistics.realtime.a_metric', datagrams: datagrams)
assert_statsd_distribution('twilio.statistics.realtime.another_metric', datagrams: datagrams)
# ...
assert_statsd_distribution('twilio.statistics.realtime.yet_another_metric', datagrams: datagrams)

# Make sure this test covers all the relevant emitted metrics
assert_no_statsd_calls(/^twilio\.statistics\.realtime\./, datagrams: datagrams)
```

I'm currently doing `assert_no_statsd_calls(datagrams: datagrams)`, but that seems too restrictive, and subject to unrelated failures. For example, if we were to start emitting metrics around the API calls we make to Twilio.

I could use `assert_no_statsd_calls('...', datagrams: datagrams)`, but that would effectively mean listing every metric with that prefix, and need to be updated any time another metric is added, to ensure this test covers all relevant metrics.

wvanbergen:
This can be done fairly cleanly after capturing the datagrams `assert datagrams.none? { |datagram| datagram.name.match?(/^twilio\.statistics\.realtime\./) }`

sambostock:
Another alternative would be to start supporting a wildcard, so you could do something like

```ruby
assert_no_statsd_calls('twilio.statistics.realtime.*')
```

sambostock:
> This can be done fairly cleanly after capturing the datagrams `assert datagrams.none? { |datagram| datagram.name.match?(/^twilio\.statistics\.realtime\./) }`

Ahh, good point. I feel like matching by prefix (not the `STATSD_PREFIX`, but the first parts of the metric) wouldn't be a bad middle ground, but that's just me.

Do you think the variadic arguments make sense, or should I close this all together?

wvanbergen:
I just ended up doing a big pass of removing features that got added over time to serve minor use cases that made maintenance of the library harder in the long run. So I rather not start re-adding features like this unless we have some evidence that this is a very common use case.

Moreover, I specifically want to push back against features that make it easier to deal with large amount of different metrics or tags, or merely by their existence "legitimizes" using a high number of metrics or tags. We have a problem of having way too many metrics. That costs us a lot of money, makes metrics less discoverable, and makes the UI slower. I would like this library to nudge users in the right direction of using as little metrics/tags as possible by having to be more precies & specific in their metric usage. 

If you want, you can still do this with the existing helper methods with the oneliner I gave you, or write your own helper method in your app. But do you really need to emit all of those metrics?

wvanbergen:
I am OK with the variadic arguments, because it still requires you to be precise

sambostock:
> I am OK with the variadic arguments, because it still requires you to be precise

Cool. I'll shrink this PR down to just that. 👌

> removing features that got added over time to serve minor use cases that made maintenance of the library harder in the long run

Fair 👍

> really need to emit all of those metrics?

We're trying to replicate a dashboard that exists inside one of Twilio's products, which requires a license to access. Moreover, having the data in Datadog allows us to build dashboard which surface the queues across the various tools, since not everything is in Twilio.

sambostock:
@wvanbergen I've updated the code and PR description to remove the RegExp support, and leave variadic arguments. If approved, I will squash the commits together before merging.

test/assertions_test.rb
```diff
@@ -33,6 +33,24 @@ def test_assert_no_statsd_calls
...
+      StatsD.increment('counter')
+    end
+
+    assertion = assert_raises(Minitest::Assertion) do
+      @test_case.assert_no_statsd_calls('counter1', 'counter2') do
+        StatsD.increment('counter0')
+        StatsD.increment('counter1')
+        StatsD.increment('counter2')
+        StatsD.increment('counter3')
+      end
```
wvanbergen: Maybe a test where `assert_no_statsd_calls` has multiple metrics, but the block only emits one of them would be a good test to add that the logic is `any`, rather than `all`.

PR #236, opened 2019-10-25
Fix lots of typos

sambostock:
This fixes most (if not all) typos in the codebase.

I found them using the [`scspell`](https://github.com/myint/scspell) tool by running

```bash
git ls-files | xargs -o scspell
```

and dealing with each unknown word, one by one. 😅 

Most of the typos were in text files, or in comments. A couple were actual code (e.g. variable names), but those should have had every occurrence replaced.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,7 +16,7 @@ _Nothing yet_
   However, if you are, you can capture StatsD datagrams using the
   `capture_statsd_datagrams` method, and run your own assertions on the list.
 - ⚠️ Remove `StatsD.client`. This was added in version 2.6.0 in order to
-  experiment with the new client. However, at this point thereare better ways
+  experiment with the new client. However, at this point there are better ways
   to do this.
   - You can set `StatsD.singleton_client` to a new client, which causes the
     calls to the StatsD singleton to be handled by a new client. If you set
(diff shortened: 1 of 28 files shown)


sambostock:
The test suite appears to be failing on Windows because of

```
Run gem install bundler && bundle install --jobs 4 --retry 3
At D:\a\_temp\48aedf55-2f9e-41d7-893b-70269dfa3633.ps1:2 char:21
+ gem install bundler && bundle install --jobs 4 --retry 3
+                     ~~
The token '&&' is not a valid statement separator in this version.
+ CategoryInfo          : ParserError: (:) [], ParseException
+ FullyQualifiedErrorId : InvalidEndOfLine
```
 which seems unrelated. I tried reverting my commit, and it still fails.

----

I also noticed that after resetting back to my original commit and force pushing, it ran GitHub ran the workflows against the commit **again**, so there are now duplicates, which I assume is a bug in GitHub workflows.

HereComeTheWalrus is also having a hard time with this page, and constantly refreshing, though I'm not sure if that's related, or just my machine being messed up 😅

wvanbergen:
Yeah, I am not sure why that started failing. If you want to, can you spin up a PR to disable this?

PR #237, opened 2019-10-25
Return truish void from metric methods

wvanbergen:
We broke backwards compatibility with the previous client by not returning metric objects when calling StatsD metric methods anymore. However, in some cases code depends on the return value of these methods to evaluate to true, because they are used as last statement in a block, which end sup being evaluated as an expression in `if` statements. It is basically impossible to detect people depending on this return value being evaluated to true.

So rather than returning `nil`, I propose we return a useless object that evaluates to true so not introduce accidental bugs by migrating to the new version.


lib/statsd/instrument/client.rb
```diff
@@ -16,6 +16,9 @@
 #   next major release of this library. While this class may already be functional,
 #   we provide no guarantees about the API and the behavior may change.
 class StatsD::Instrument::Client
+  VOID = Object.new.freeze
```
casperisfine: Might be a bit overkill, but maybe a BasicObject subclass would give a better error message:

```ruby
>> DeprectedReturnValue = Class.new(BasicObject)
=> DeprectedReturnValue
>> VOID = DeprectedReturnValue.new
(Object doesn't support #inspect)
=> 
>> VOID.foo
Traceback (most recent call last):
        4: from /opt/rubies/2.6.5/bin/irb:23:in `<main>'
        3: from /opt/rubies/2.6.5/bin/irb:23:in `load'
        2: from /opt/rubies/2.6.5/lib/ruby/gems/2.6.0/gems/irb-1.0.0/exe/irb:11:in `<top (required)>'
        1: from (irb):3
NoMethodError (undefined method `foo' for #<DeprectedReturnValue:0x00007f82a8968a48>)
```

vs

```ruby
>> Object.new.freeze.plop
Traceback (most recent call last):
        5: from /opt/rubies/2.6.5/bin/irb:23:in `<main>'
        4: from /opt/rubies/2.6.5/bin/irb:23:in `load'
        3: from /opt/rubies/2.6.5/lib/ruby/gems/2.6.0/gems/irb-1.0.0/exe/irb:11:in `<top (required)>'
        2: from (irb):4
        1: from (irb):4:in `rescue in irb_binding'
NoMethodError (undefined method `plop' for #<Object:0x00007f82a895b028>)
```

wvanbergen: I think I will do this because why not? But I am not super worried about people actually using the return value in this way. It's really the `if` case that worries me.

PR #238, opened 2019-10-25
Disable Windows CI check

sambostock:
This disables the CI check against Windows, as per [@wvanbergen's comment on #236](https://github.com/Shopify/statsd-instrument/pull/236#issuecomment-546405005).

For some reason, the Windows check started failing with the following error.

```
Run gem install bundler && bundle install --jobs 4 --retry 3
At D:\a\_temp\48aedf55-2f9e-41d7-893b-70269dfa3633.ps1:2 char:21
+ gem install bundler && bundle install --jobs 4 --retry 3
+                     ~~
The token '&&' is not a valid statement separator in this version.
+ CategoryInfo          : ParserError: (:) [], ParseException
+ FullyQualifiedErrorId : InvalidEndOfLine
```

Temporarily excluding Windows unblocks CI for other PRs until this is figured out.

----
Closes #239

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -13,15 +13,6 @@ jobs:
         ruby: [2.3, 2.4, 2.5, 2.6]
 
         exclude:
-        # Windows started failing on the following command for some reason.
-        #
-        #     gem install bundler && bundle install --jobs 4 --retry 3
-        #                         ~~
-        #     The token '&&' is not a valid statement separator in this version.
-        #
-        # Disabling windows builds for the time being.
-        - platform: windows-2019
-
         # The Windows environment does not support older Ruby versions. We only test against the latest version
         - platform: windows-2019
           ruby: 2.3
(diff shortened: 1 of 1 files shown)


sambostock:
I may have just found a bug in GitHub Actions?

I originally tried

```yaml
exclude:
  - platform: windows-2019
  # ...
```

But it looks like you need to [exclude specific combinations](https://help.github.com/en/github/automating-your-workflow-with-github-actions/workflow-syntax-for-github-actions#example-excluding-configurations-from-a-matrix), so I changed it to

```yaml
exclude:
  - platform: windows-2019
    ruby: 2.6
  # ...
```

Looks like now the check list expects `Ruby 2.6 on windows-2019`, because of my first commit, but it will never run cause my second commit excludes it. 🤔 

image

sambostock:
Still not working. Tried opening #238 as a new PR, but that didn't work either. Any ideas @wvanbergen? Is there somewhere where that check is configured as required? Is GitHub Actions using the config from `master` instead of this branch?

wvanbergen:
I had to update branch protection rules.

PR #245, opened 2019-10-31
Client#measure: raise ArgumentError if neither a block or value is provided

wvanbergen:
This fixes a small compatibility issue with the legacy client. This also applies to `Client#distribition`

### DIFF
--- a/lib/statsd/instrument/client.rb
+++ b/lib/statsd/instrument/client.rb
@@ -120,6 +120,8 @@ class StatsD::Instrument::Client
   def measure(name, value = nil, sample_rate: nil, tags: nil, no_prefix: false, &block)
     if block_given?
       return latency(name, sample_rate: sample_rate, tags: tags, metric_type: :ms, no_prefix: no_prefix, &block)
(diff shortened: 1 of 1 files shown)


lib/statsd/instrument/client.rb
```diff
@@ -120,6 +120,8 @@ def increment(name, value = 1, sample_rate: nil, tags: nil, no_prefix: false)
   def measure(name, value = nil, sample_rate: nil, tags: nil, no_prefix: false, &block)
     if block_given?
       return latency(name, sample_rate: sample_rate, tags: tags, metric_type: :ms, no_prefix: no_prefix, &block)
+    elsif value.nil?
```
Edouard-chin: Should we raise an error to avoid confusion if a block and a value is given ?

wvanbergen:
Actually, I think I will remove this from the regular client, and move this to "strict mode" instead. It's kind of pointless to check this for every StatsD metric in production and pay the performance penalty. Strict mode exists to find these kinds of things.

PR #246, opened 2019-10-31
Version 3.0 prep

wvanbergen:
This removes the old client and all associated files, and fixes up the test suite to no longer depend on it.

My plan was to release a beta vewrsion of 3.0 based on this branch, and try to update to it in Shopify. We'll likely have to fix up the dependency requirements of some other gems as well in order for bundler to be happy.


lib/statsd/instrument/helpers.rb
```diff
@@ -3,38 +3,9 @@
 module StatsD::Instrument::Helpers
   def capture_statsd_datagrams(client: nil, &block)
     client ||= StatsD.singleton_client
```
casperisfine: `capture_statsd_datagrams(client: StatsD.singleton_client, &block)` would make more sense IMO.

wvanbergen: That doesn't work, unless we propagate this default value out to all the methods that call this. Right now, if I do that tests will fail because we call `capture` on `nil`.

PR #247, opened 2019-10-31
Make void a regular but frozen object, rather than a basic object

wvanbergen:
Using a `BasicObject` for `VOID` return values is a bit too restrictive. Many things that work on object are not supported (e.g. `#tap` and `#class`), which breaks all kinds of instrumentation wrapper methods. Technically that means that those methods depend on the return value of a StatsD metric and should be fixed, but it creates a bit too much pain.

### DIFF
--- a/lib/statsd/instrument.rb
+++ b/lib/statsd/instrument.rb
@@ -344,9 +344,9 @@ module StatsD
       remove_from_method(method, name, :distribution)
     end
 
-    VoidClass = Class.new(BasicObject)
+    VoidClass = Class.new
     private_constant :VoidClass
-    VOID = VoidClass.new
+    VOID = VoidClass.new.freeze
 
     private
 


lib/statsd/instrument.rb
```diff
@@ -344,9 +344,9 @@ def statsd_remove_distribution(method, name)
       remove_from_method(method, name, :distribution)
     end
 
-    VoidClass = Class.new(BasicObject)
+    VoidClass = Class.new
```
Edouard-chin: Why not just `Object.new?`

lib/statsd/instrument.rb
```diff
@@ -344,9 +344,9 @@ def statsd_remove_distribution(method, name)
...
     private_constant :VoidClass
-    VOID = VoidClass.new
+    VOID = VoidClass.new.freeze
```
rafaelfranca: Will freezing this object does anything? I don't think it has a state, so there is nothing you can do to change that state.

lib/statsd/instrument.rb
wvanbergen: This gives you a more targeted error message ("`foo` is not a method on `VoidClass`" rather than "`foo` is not a method on `Object`").

lib/statsd/instrument.rb
wvanbergen: Not really, I did this more as "best practice" for singleton objects that should not change

PR #248, opened 2019-11-06
Document assert_statsd_increment with value

agarcher:
I spent a while trying to figure out how to assert a call to `increment` with a non-default step size (aka `value`). The implementation is not obvious because `StatsD.increment` uses a positional argument, while `assert_statsd_increment` uses a named argument. Documentation of this seems useful to me.

### DIFF
--- a/README.md
+++ b/README.md
@@ -269,6 +269,13 @@ class MyTestcase < Minitest::Test
       StatsD.increment('counter.name', sample_rate: 1.0) # matches
       StatsD.increment('counter.name', sample_rate: 0.1) # matches too
     end
+    
+    # Set `value` if incrementing by more than 1:
+    assert_statsd_increment('counter.name', value: 5) do
+      StatsD.increment('unrelated', 5) # doesn't match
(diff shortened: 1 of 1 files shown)


wvanbergen:
This is a leftover from a bunch of refactoring We should really fix this for version 3.0, so we accept a positional argument.

agarcher:
Yes, accepting a positional argument would be way better. Makes sense to me. So you want to just close this and wait for 3.0?

wvanbergen:
How I dealt with the other deprecations:
- deprecate on 2.X
- remove in 3.0 release

However, for some deprecations in the test helpers I simply immediately removed stuff in a 2.x version, especially when we can give a good error message and the fix is simple. I am leaning towards this option here.

wvanbergen:
https://github.com/Shopify/statsd-instrument/pull/249

wvanbergen:
Closing this. 2.9.2 has been released, which allow you to use a positional argument instead.

agarcher:
Thanks @wvanbergen!

PR #252, opened 2019-12-01
Update Readme to correct default settings

mike-stewart:
The documentation says that the default implementation is `statsd`, but it appears to actually be `datadog`. This updates the README.

https://github.com/Shopify/statsd-instrument/blob/02e682d463eba0043db1935d15c0ab1b523e9dd5/lib/statsd/instrument/environment.rb#L61


wvanbergen:
Fixed by #270

PR #254, opened 2020-01-06
Bump to version 3.0.0

wvanbergen:
This bumps the version to 3.0.0, a major release in which we remove the legacy client and all deprecated functionality. There were 2 pre-releases and no new problems came up, so I think we are good to do this now.

I added basic upgrade instructions to the CHANGELOG.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,7 +10,7 @@ _Nothing yet_
 
 ## Version 3.0.0
 
-This version makes the new client dat was added in version 2.6+ the default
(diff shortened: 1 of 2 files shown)


CHANGELOG.md
```diff
@@ -8,6 +8,29 @@ section below.
 
 _Nothing yet_
 
+## Version 3.0.0
+
+This version makes the new client dat was added in version 2.6+ the default
```
dylanahsmith:
```suggestion
This version makes the new client that was added in version 2.6+ the default
```

PR #257, opened 2020-03-17
Let the UDP sink use a dispatcher thread

methodmissing:
### Why?

Although UDP uses a connectionless communication model, the following user space code paths aren't free:

* [Mutex](https://github.com/ruby/ruby/blob/115fec062ccf7c6d72c8d5f64b7a5d84c9fb2dd8/thread_sync.c#L237-L317) held on the socket
* Sending datagrams still incurs a [system call](https://github.com/ruby/ruby/blob/1663d347c993debf7ed83e11e291e7a21e14ed03/ext/socket/udpsocket.c#L199-L220)

Using a `Queue` as contract with a simple dispatch Thread reduce datagram dispatch for the producer to an [array push](https://github.com/ruby/ruby/blob/115fec062ccf7c6d72c8d5f64b7a5d84c9fb2dd8/thread_sync.c#L784) and a user space [thread interrupt](https://github.com/ruby/ruby/blob/0256e4f0f5e10f0a15cbba2cd64e252dfa864e4a/vm_core.h#L1791 ) 

Master:

```
methodmissing:statsd-instrument lourens$ benchmark/send-metrics-to-local-udp-receiver
StatsD metrics to local UDP receiver (branch: master, sha: 07fb158)
                         1.213k i/100ms
StatsD metrics to local UDP receiver (branch: master, sha: 07fb158)
                         11.876k (± 3.9%) i/s -     59.437k in   5.012616s

To compare the performance of this revision against another revision (e.g. master),
check out a different branch and run this benchmark script again.
```

This PR:

```
methodmissing:statsd-instrument lourens$ benchmark/send-metrics-to-local-udp-receiver
StatsD metrics to local UDP receiver (branch: dispatch-thread, sha: 3137f7e)
                         4.199k i/100ms
StatsD metrics to local UDP receiver (branch: dispatch-thread, sha: 3137f7e)
                         44.251k (±25.9%) i/s -    201.552k in   5.042619s

Comparison:
StatsD metrics to local UDP receiver (branch: dispatch-thread, sha: 3137f7e):    44251.3 i/s
StatsD metrics to local UDP receiver (branch: master, sha: 07fb158):    11921.7 i/s - 3.71x  (± 0.00) slower
```

### Why not?

Threads 😄 The dispatcher is very thin and cleanly separates IO and socket init + teardown on error from the metrics producer thread. Cleanup is handled with a GC finalizer (which of course only fires on clean exit, but this is true for ObjectSpace in general too). See the throwaway section below

### Why this is throwaway

Queues do not work X process because https://github.com/ruby/ruby/blob/115fec062ccf7c6d72c8d5f64b7a5d84c9fb2dd8/thread_sync.c#L606-L619 and therefore test https://github.com/Shopify/statsd-instrument/blob/master/test/udp_sink_test.rb#L63-L82 would always fail because https://github.com/Shopify/statsd-instrument/blob/master/test/udp_sink_test.rb#L67 is enq'ed but `num_waiting` is reset to 0, with the dispatcher thread effectively never seeing the datagram as the queue has no waiters.

### DIFF
--- a/lib/statsd/instrument/udp_sink.rb
+++ b/lib/statsd/instrument/udp_sink.rb
@@ -1,5 +1,8 @@
 # frozen_string_literal: true
 
+require 'objspace'
+require 'thread'
+
 # @note This class is part of the new Client implementation that is intended
 #   to become the new default in the next major release of this library.
 class StatsD::Instrument::UDPSink
@@ -8,13 +11,62 @@ class StatsD::Instrument::UDPSink
     new(host, Integer(port_as_string))
   end
 
+  class Dispatcher
+    def initialize(host, port, queue)
+      @host = host
+      @port = port
+      @queue = queue
+      @socket = nil
+      @thread = nil
+      ObjectSpace.define_finalizer(self, Proc.new { |o| o.stop })
+    end
+
+    def start
+      @thread = Thread.new { dispatcher_loop }
+      self
+    end
+
+    def stop
+      @queue.close
+      @thread.join(0.1)
+    rescue ThreadError
+    end
+
+    private
+    def dispatcher_loop
+      loop do
+        datagram = @queue.pop(false)
+        begin
+          socket.send(datagram, 0)
+        rescue ThreadError
+          socket.send(datagram, 0)
+        rescue SocketError, IOError, SystemCallError
+          # TODO: log?
+          invalidate_socket
+        end
+      end
+    end
+
+    def socket
+      if @socket.nil?
+        @socket = UDPSocket.new
+        @socket.connect(@host, @port)
+      end
+      @socket
+    end
+
+    def invalidate_socket
+      @socket = nil
+    end
+  end
+
   attr_reader :host, :port
 
   def initialize(host, port)
     @host = host
     @port = port
-    @mutex = Mutex.new
-    @socket = nil
+    @queue = Queue.new
+    @dispatcher = Dispatcher.new(host, port, @queue).start
   end
 
   def sample?(sample_rate)
@@ -22,41 +74,11 @@ class StatsD::Instrument::UDPSink
   end
 
   def <<(datagram)
-    with_socket { |socket| socket.send(datagram, 0) > 0 }
+    @queue << datagram
     self
-
-  rescue ThreadError
-    # In cases where a TERM or KILL signal has been sent, and we send stats as
-    # part of a signal handler, locks cannot be acquired, so we do our best
-    # to try and send the datagram without a lock.
-    socket.send(datagram, 0) > 0
-
-  rescue SocketError, IOError, SystemCallError
-    # TODO: log?
-    invalidate_socket
   end
 
   def addr
     "#{host}:#{port}"
-  end
-
-  private
-
-  def with_socket
-    @mutex.synchronize { yield(socket) }
-  end
-
-  def socket
-    if @socket.nil?
-      @socket = UDPSocket.new
-      @socket.connect(@host, @port)
-    end
-    @socket
-  end
-
-  def invalidate_socket
-    @mutex.synchronize do
-      @socket = nil
-    end
   end
 end
--- a/test/udp_sink_test.rb
+++ b/test/udp_sink_test.rb
@@ -58,6 +58,8 @@ class UDPSinkTest < Minitest::Test
     udp_sink = StatsD::Instrument::UDPSink.new('localhost', 8125)
     udp_sink << 'foo:1|c'
     udp_sink << 'bar:1|c'
+    # Let the dispatcher thread emit
+    sleep 0.1
   end
 
   def test_sends_datagram_in_signal_handler


lib/statsd/instrument/udp_sink.rb
```diff
@@ -8,55 +11,74 @@ def self.for_addr(addr)
...
   end
 
+  class Dispatcher
+    def initialize(host, port, queue)
+      @host = host
+      @port = port
+      @queue = queue
+      @socket = nil
+      @thread = nil
+      ObjectSpace.define_finalizer(self, Proc.new { |o| o.stop })
```
kirs: For my own education, when would you choose to use `at_exit` and when `ObjectSpace.define_finalizer`?

lib/statsd/instrument/udp_sink.rb
```diff
@@ -8,55 +11,74 @@ def self.for_addr(addr)
...
+  end
+
   attr_reader :host, :port
 
   def initialize(host, port)
     @host = host
     @port = port
-    @mutex = Mutex.new
-    @socket = nil
+    @queue = Queue.new
```
wvanbergen: Should this queue be bounded in some way?

test/udp_sink_test.rb
```diff
@@ -58,6 +58,8 @@ def test_socket_error_should_invalidate_socket
     udp_sink = StatsD::Instrument::UDPSink.new('localhost', 8125)
     udp_sink << 'foo:1|c'
     udp_sink << 'bar:1|c'
+    # Let the dispatcher thread emit
+    sleep 0.1
```
wvanbergen: would it be nice to have a `udp_sunk.flush` or another method that blocks until the queue is empty?

wvanbergen:
What would be the impact of this when using a forking web/job server?

casperisfine:
I love the idea. However I wonder if it would be worth having a single handler rather than one per udpsink.

But maybe that's not worth it as it's unlikelt to have multiple sink instances at the same time?

lib/statsd/instrument/udp_sink.rb
casperisfine: `at_exit` is called when the process exists.

`define_finalizer` is called when the object is garbaged collected.

Here a finalizer is better, because if you dereference a sink, you want it to be cleaned up immediately rather than on the process exit.

casperisfine: Maybe use `:stop.to_proc` to ensure you're not enclosing anything in the proc scope? I don't think it's a problem right now, but at least `to_proc` would make it more clear.

csfrancis:
> The Queue instance would still involve a mutex / synchronization point, right? So the perf difference comes from making the syscall async.

No, there's no mutex. The difference is that for MRI, the GVL is preventing multiple threads from running at the same time. The `Queue#push` will trigger an MRI interrupt, which wakes up a thread that is blocking via `Queue#pop`: https://github.com/ruby/ruby/blob/115fec062ccf7c6d72c8d5f64b7a5d84c9fb2dd8/thread_sync.c#L785. This all happens in user-space without a syscall.

lib/statsd/instrument/udp_sink.rb
```diff
@@ -8,55 +11,74 @@ def self.for_addr(addr)
...
+    def stop
+      @queue.close
+      @thread.join(0.1)
+    rescue ThreadError
+    end
+
+    private
+    def dispatcher_loop
+      loop do
+        datagram = @queue.pop(false)
```
casperisfine: `false` is the default argument here. I'd would simply not pass it.

lib/statsd/instrument/udp_sink.rb
```diff
@@ -8,55 +11,74 @@ def self.for_addr(addr)
...
+        begin
+          socket.send(datagram, 0)
+        rescue ThreadError
+          socket.send(datagram, 0)
```
casperisfine: How is `UDPSocket#send` expected to raise a `ThreadError` and why would you retry it?

lib/statsd/instrument/udp_sink.rb
casperisfine: Also you need to check `datagram` isn't `nil`:

```ruby
>> q = Queue.new
=> #<Thread::Queue:0x00007fdf7d81ed68>
>> q.close
=> #<Thread::Queue:0x00007fdf7d81ed68>
>> q.pop
=> nil
```

lib/statsd/instrument/udp_sink.rb
csfrancis: I'm assuming this was copypasta from https://github.com/Shopify/statsd-instrument/pull/257/files#diff-94686ddb5f351601e4da59e66916633fL29-L31. But I don't think it's necessary anymore now that there's no locking.

dylanahsmith:
> No, there's no mutex. [...]. This all happens in user-space without a syscall.

If another thread is blocked on the queue, then it will take a mutex to interrupt the other thread in  [rb_threadptr_interrupt_common](https://github.com/ruby/ruby/blob/v2_7_0/thread.c#L491-L507).  Also, I would expect that waking up another native thread would also require another system call in the unblock function call `(th->unblock.func)(th->unblock.arg)`.  So it looks like this only gets to all happen in user-space when the queue isn't already empty.

I'm concerned the benchmark results aren't representative of real use of this library, since it seems like the performance benefit comes from queuing multiple datagrams consecutively.  If datagrams are always queued to an empty queue, then it seems like this could add extra overhead.

> I think a future improvement could be to buffer some packets and flush them out as batch

This seems like it could be a much better approach.  For instance, we could buffer datagrams during a unit of work (e.g. web request) and use [sendmmsg](https://linux.die.net/man/2/sendmmsg) to flush the buffer at the end of that unit of work or when the buffer gets full in the middle of that unit of work.

wvanbergen:
The DogStatsD library can buffer metrics in a single UDP datagram by separating them with newlines: https://github.com/DataDog/dogstatsd-ruby/blob/master/lib/datadog/statsd/batch.rb

wvanbergen:
@dylanahsmith If we deploy this branch to production, what metrics should we track to knowhow much of an impact this change has? Regular profiling?

dylanahsmith:
I would look at whatever stats we were looking at that led us to do this optimization.  It sounds like we found this hotspot through profiling, so I think it makes sense to compare the profile from after this to one we captured before.  Also, if we believe this was related to a performance regression that we noticed and we were hoping this would help resolve, then we could see if this helps with that metric.

Do we plan on reverting this if it has a negligible effect?

casperisfine:
I took the liberty to rebase this and to add a few fixes. Notably:

### Fork handling

When forking all the thread but the main one are dead in the child process. So if we use a dispatcher thread, they need to be respawned after fork. Either automatically (e..g by checking `Process.pid` all the time, or [by decorating `Process.fork`](https://github.com/rails/rails/blob/5de2dbff16009ebf8441e259d95d4b22d068ac8c/activesupport/lib/active_support/fork_tracker.rb)) or either by requiring the users to call some method.

I went for the later and introduced a `.after_fork` method for that purpose. But I'm open to implement "fork tracking" too. It would be easier to use, but libraries monkey patching core methods are not generally well received. 

It's also important to clear the queues after fork, because any message left in them would have been inserted in the parent process, so we don't want to duplicate them.

### Make it optional

Instead of changing the existing `UPDSink` I added an extra `ThreadedUDPSink`, I feel like it's better to have it as an option at least until we experiment with it in production. We cal always make it the default later on.

### Tests

I refactored the tests a bit to be more reliable. Mostly the one about sending events from a signal handler. Since the mutex is eliminated, it's no longer a problem. However the if forked process exit too quickly after the packet is written, it won't make it to the parent. 

@methodmissing @wvanbergen @dylanahsmith what do you think?

lib/statsd/instrument/threaded_udp_sink.rb
```diff
@@ -0,0 +1,110 @@
...
+          end
+        end
+
+        def initialize(host, port, queue)
+          @host = host
+          @port = port
+          @queue = queue
+          @socket = nil
+          @thread = nil
+          ObjectSpace.define_finalizer(self, self.class.finalizer(self))
```
casperisfine, line 26: Oh, and there was also an issue with the finalizer, it would never fire because of that common gotcha: https://www.mikeperham.com/2010/02/24/the-trouble-with-ruby-finalizers/

casperisfine:
> makes sense to compare the profile from after this to one we captured before.

The problem with this is that the overhead isn't really reduced, just moved in a background thread. I think a profiler would probably report a very similar time spent in the sink.

PR #259, opened 2020-03-30
Datagram builder as a C extension

methodmissing:
### Why?

The datagram builder is heavy on allocation (and reallocation) and coercing both metric names and tags into normalized and valid statsd wire protocol components.

This extension is also implemented as a wrapped struct without any global state as @hkdsun and @bmansoob expressed interest in how to do that.

* A [simple struct](https://github.com/Shopify/statsd-instrument/compare/datagram-builder-c-extension?expand=1#diff-703d6028f3c214b5fe6ab8682a33195eR15-R31) with mixed native type and `VALUE` (Ruby object reference) members

* GC [mark callback](https://github.com/Shopify/statsd-instrument/compare/datagram-builder-c-extension?expand=1#diff-703d6028f3c214b5fe6ab8682a33195eR33-R48) invoked during the GC tracing phase (we need to let the GC know about Ruby object references we hold on to)

* GC [free callback](https://github.com/Shopify/statsd-instrument/compare/datagram-builder-c-extension?expand=1#diff-703d6028f3c214b5fe6ab8682a33195eR50-R64) invoked during the sweep phase if this object was deemed to not be referenced by any other, the stack etc. . We free the symbol tables for caches and the struct itself.

* It's a good practice to add an [object size accumulator callback](https://github.com/Shopify/statsd-instrument/compare/datagram-builder-c-extension?expand=1#diff-703d6028f3c214b5fe6ab8682a33195eR66-R81) to accurately reflect the size of this object via `Objspace.memsize_of`

* A [data type declaration](https://github.com/Shopify/statsd-instrument/compare/datagram-builder-c-extension?expand=1#diff-703d6028f3c214b5fe6ab8682a33195eR83-R92) defines the shape of the wrapped object and callbacks for the GC

* An [unboxing function](https://github.com/Shopify/statsd-instrument/compare/datagram-builder-c-extension?expand=1#diff-703d6028f3c214b5fe6ab8682a33195eR94) coerces a `VALUE` reference back to a builder struct as per the type definition above

* The [allocator function](https://github.com/Shopify/statsd-instrument/compare/datagram-builder-c-extension?expand=1#diff-703d6028f3c214b5fe6ab8682a33195eR98-R111) prepares the struct for use by initializing members to their appropriate types, caches conditionally etc. and returns a `VALUE` reference to the allocated struct on the heap

### Points of attack

* The prefix is constant, initialized once yet [merged into the output buffer from offset 0](https://github.com/Shopify/statsd-instrument/blob/master/lib/statsd/instrument/datagram_builder.rb#L96) to it's length for every builder call. We can instead [seed](https://github.com/Shopify/statsd-instrument/compare/datagram-builder-c-extension?expand=1#diff-703d6028f3c214b5fe6ab8682a33195eR121-R127) the builder buffer with it and [advance the buffer low water mark](https://github.com/Shopify/statsd-instrument/compare/datagram-builder-c-extension?expand=1#diff-703d6028f3c214b5fe6ab8682a33195eR242) to the end of prefix  on every call

* [Name normalization](https://github.com/Shopify/statsd-instrument/blob/master/lib/statsd/instrument/datagram_builder.rb#L88-L92) is expensive as it involves a regex match for the fast path and an additional function call for the slow path. For the fast path it's more efficient to [walk to string from start to end](https://github.com/Shopify/statsd-instrument/compare/datagram-builder-c-extension?expand=1#diff-703d6028f3c214b5fe6ab8682a33195eR135-R156url) , and for the slow path implement a [bounded size normalized names cache](https://github.com/Shopify/statsd-instrument/compare/datagram-builder-c-extension?expand=1#diff-703d6028f3c214b5fe6ab8682a33195eR165-R186) to save on `String#tr` function calls for invalid metric names.

* [Tags normalization](https://github.com/Shopify/statsd-instrument/blob/master/lib/statsd/instrument/datagram_builder.rb#L78) is expensive and especially with the hybrid `Hash` and `Array` tags API allocates additional objects for the `Hash` based API. This method is not optimized in C as it's mostly iterators and method calls which can become EXTREMELY ugly rewriting in a C extension. Instead another [bounded size normalized tags cache](https://github.com/Shopify/statsd-instrument/compare/datagram-builder-c-extension?expand=1#diff-703d6028f3c214b5fe6ab8682a33195eR188%60-R211) was introduced to bypass this Ruby method for tags collections with a contents hash we've seen before.

* There's always a [new tags Array spawned ](https://github.com/Shopify/statsd-instrument/blob/master/lib/statsd/instrument/datagram_builder.rb#L95) in the `generate_generic_datagram` method. This is wasteful even for cases which don't have any default tags defined. This was flattened out into a [zero alloc append only](https://github.com/Shopify/statsd-instrument/compare/datagram-builder-c-extension?expand=1#diff-703d6028f3c214b5fe6ab8682a33195eR292-R307) pattern instead.

* Sample rate values appended to the buffer [is zero alloc for Fixnum and Float types](https://github.com/Shopify/statsd-instrument/compare/datagram-builder-c-extension?expand=1#diff-703d6028f3c214b5fe6ab8682a33195eR269-R290), but has a fallback path for other object types which would allocate a ruby String. This pattern likely works well for the `value` argument too, but would prefer to get runtime feedback first and consider it as a later optimisation.

* Remove a repeated ivar lookup and size check for default tags in favor of [caching it on the builder struct](https://github.com/Shopify/statsd-instrument/compare/datagram-builder-c-extension?expand=1#diff-703d6028f3c214b5fe6ab8682a33195eR129-R131) instead on init.

* Reduce method dispatch overhead by [calling the builder directly](https://github.com/Shopify/statsd-instrument/compare/datagram-builder-c-extension?expand=1#diff-703d6028f3c214b5fe6ab8682a33195eR314-R347) from the various metric helper methods.

### Configurables

* The buffer size can be changed [compile time](https://github.com/Shopify/statsd-instrument/compare/datagram-builder-c-extension?expand=1#diff-703d6028f3c214b5fe6ab8682a33195eR8-R11)

* Normalized caches are compile time features that can easily be [disabled or changed](https://github.com/Shopify/statsd-instrument/compare/datagram-builder-c-extension?expand=1#diff-703d6028f3c214b5fe6ab8682a33195eR8-R11)

### Resiliency issues covered

* Strict [overflow checks](https://github.com/Shopify/statsd-instrument/compare/datagram-builder-c-extension?expand=1#diff-703d6028f3c214b5fe6ab8682a33195eR286) that finalize the buffer prior to overflow

* Test suite run with [GC.stress = true](https://github.com/Shopify/statsd-instrument/compare/datagram-builder-c-extension?expand=1#diff-f14ca9959ff06f3ca2ee6b9564866954R20)

### Wrapped builder struct layout

Fits within 1 cache line (46 of 64 bytes) with the buffer as the last member. 4kb is very generous, but in reality the vast majority of statsd datagrams would only reach into 1 or 2 more cache lines.

The struct is passed around by reference as it's heap allocated anyways and few things are as bad as large structs passed by value.

```
lourens@CarbonX1:~/src/statsd-instrument/lib/statsd/instrument/ext$ pahole -C datagram_builder ./statsd.so
struct datagram_builder {
	st_table *                 normalized_tags_cache; /*     0     8 */
	st_table *                 normalized_names_cache; /*     8     8 */
	VALUE                      str_normalize_chars;  /*    16     8 */
	VALUE                      str_normalize_replacement; /*    24     8 */
	VALUE                      default_tags;         /*    32     8 */
	_Bool                      empty_default_tags;   /*    40     1 */

	/* XXX 3 bytes hole, try to pack */

	int                        prefix_len;           /*    44     4 */
	int                        len;                  /*    48     4 */
	char                       datagram[4096];       /*    52  4096 */

	/* size: 4152, cachelines: 65, members: 9 */
	/* sum members: 4145, holes: 1, sum holes: 3 */
	/* padding: 4 */
	/* last cacheline: 56 bytes */
};
```

### Future TODO

* Remove the String allocation for the `value` argument
* Evaluate the [hashing function](https://github.com/Shopify/statsd-instrument/compare/datagram-builder-c-extension?expand=1#diff-703d6028f3c214b5fe6ab8682a33195eR198) used for the tags cache (Hash and Array specific are not directly exposed through the extension API)

### Downsides

* This library now becomes dependent of a C compiler and ruby development headers - need to think about decoupling so that is optional.

### Other ideas

* Willem mentioned more Stricter mode linters that can catch invalid metric names and tags enabled for example during CI could work wonders too.

### DIFF
--- a/ext/statsd/statsd.c
+++ b/ext/statsd/statsd.c
@@ -1,5 +1,6 @@
 #include <ruby.h>
 #include <ruby/encoding.h>
+#include <ruby/st.h>
 
 #define DATAGRAM_SIZE_MAX 4096
 #define NORMALIZED_TAGS_CACHE_ENABLED 1
@@ -7,15 +8,117 @@
 #define NORMALIZED_NAMES_CACHE_ENABLED 1
 #define NORMALIZED_NAMES_CACHE_MAX 512
 
-static ID idTr, idNormalizeTags, idDefaultTags, idPrefix, idNormalizedTagsCache, idNormalizedNamesCache;
+static ID idTr, idNormalizeTags, idDefaultTags, idPrefix;
 static VALUE strNormalizeChars, strNormalizeReplacement;
+
+struct datagram_builder {
+#ifdef NORMALIZED_TAGS_CACHE_ENABLED
+  st_table *normalized_tags_cache;
+#endif
+#ifdef NORMALIZED_NAMES_CACHE_ENABLED
+  st_table *normalized_names_cache;
+#endif
+  // cached default tags ivar to skip a lookup
+  VALUE default_tags;
+  long prefix_len;
+  // last member to not glob up cache lines to access other struct members
+  char datagram[DATAGRAM_SIZE_MAX];
+};
+
+// GC callback to mark the wrapper struct. Conditionally symbol tables if caching is enabled (values only)
+// and the cached default tags as well.
+void
+datagram_builder_mark(void *ptr)
+{
+  const struct datagram_builder *builder = (struct datagram_builder *)ptr;
+#ifdef NORMALIZED_TAGS_CACHE_ENABLED
+  rb_mark_tbl(builder->normalized_tags_cache);
+#endif
+#ifdef NORMALIZED_NAMES_CACHE_ENABLED
+  rb_mark_tbl(builder->normalized_names_cache);
+#endif
+  rb_gc_mark(builder->default_tags);
+}
+
+// GC callback to free the wrapper struct. Conditionally symbol tables if caching is enabled
+// and the struct itself.
+void
+datagram_builder_free(void *ptr)
+{
+  struct datagram_builder *builder = (struct datagram_builder *)ptr;
+  if (!builder) return;
+#ifdef NORMALIZED_TAGS_CACHE_ENABLED
+  st_free_table(builder->normalized_tags_cache);
+#endif
+#ifdef NORMALIZED_NAMES_CACHE_ENABLED
+  st_free_table(builder->normalized_names_cache);
+#endif
+  xfree(builder);
+}
+
+// Be nice to ObjectSpace and let the size be known. We'd likely want some feedback on
+// this with various normalized cache size values.
+size_t
+datagram_builder_size(const void *ptr)
+{
+  size_t size;
+  const struct datagram_builder *builder = (struct datagram_builder *)ptr;
+  size = sizeof(struct datagram_builder);
+#ifdef NORMALIZED_TAGS_CACHE_ENABLED
+  size += st_memsize(builder->normalized_tags_cache);
+#endif
+#ifdef NORMALIZED_NAMES_CACHE_ENABLED
+  size += st_memsize(builder->normalized_names_cache);
+#endif
+  return size;
+}
+
+const rb_data_type_t datagram_builder_type = {
+  .wrap_struct_name = "datagram_builder",
+  .function = {
+    .dmark = datagram_builder_mark,
+    .dfree = datagram_builder_free,
+    .dsize = datagram_builder_size,
+  },
+  .data = NULL,
+  .flags = RUBY_TYPED_FREE_IMMEDIATELY,
+};
+
+#define get_datagram_builder_struct(self) \
+  struct datagram_builder *builder = NULL; \
+  TypedData_Get_Struct(self, struct datagram_builder, &datagram_builder_type, builder); \
+
+static VALUE
+datagram_builder_alloc(VALUE self)
+{
+  struct datagram_builder *builder = ZALLOC(struct datagram_builder);
+#ifdef NORMALIZED_TAGS_CACHE_ENABLED
+  builder->normalized_tags_cache = st_init_numtable_with_size(NORMALIZED_TAGS_CACHE_MAX);
+#endif
+#ifdef NORMALIZED_NAMES_CACHE_ENABLED
+  builder->normalized_names_cache = st_init_numtable_with_size(NORMALIZED_NAMES_CACHE_MAX);
+#endif
+  return TypedData_Wrap_Struct(self, &datagram_builder_type, builder);
+}
 
 static VALUE
 initialize(int argc, VALUE *argv, VALUE self)
 {
+  VALUE prefix;
+  long chunk_len = 0;
+  get_datagram_builder_struct(self);
   rb_call_super(argc, argv);
-  rb_ivar_set(self, idNormalizedTagsCache, rb_obj_hide(rb_hash_new()));
-  rb_ivar_set(self, idNormalizedNamesCache, rb_obj_hide(rb_hash_new()));
+
+  // pre seed the buffer with the prefix and advance the offset as it's fixed for the lifetime of
+  // the builder
+  prefix = rb_ivar_get(self, idPrefix);
+  if ((chunk_len = RSTRING_LEN(prefix)) != 0 && chunk_len < DATAGRAM_SIZE_MAX) {
+    memcpy(builder->datagram, StringValuePtr(prefix), chunk_len);
+    builder->prefix_len = chunk_len;
+  }
+
+  // Cache the defaukt tags ivar on the lookup struct
+  builder->default_tags = rb_ivar_get(self, idDefaultTags);
   return self;
 }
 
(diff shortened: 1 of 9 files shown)


methodmissing:
... and some CI massaging required

Screenshot from 2020-03-30 22-56-46

wvanbergen:
> This library now becomes dependent of a C compiler and ruby development headers - need to think about decoupling so that is optional.

In think there are two ways to address this:
1. The datagram builder is already configurable. We could ship this as a separate datagram builder class (potentially as a different gem). The client can then configure to use this different datagram builder class, maybe using an environment variable.
2. We create a separate gem that when loaded, will overwrite the datagram builder class with this native implementation.

methodmissing:
Yeah that makes sense - like the `sequel` model

ext/statsd/statsd.c
```diff
@@ -0,0 +1,376 @@
...
+
+// GC callback to free the wrapper struct. Conditionally symbol tables if caching is enabled
+// and the struct itself.
+void
+datagram_builder_free(void *ptr)
+{
+  struct datagram_builder *builder = (struct datagram_builder *)ptr;
+  if (!builder) return;
+#ifdef NORMALIZED_TAGS_CACHE_ENABLED
+  st_free_table(builder->normalized_tags_cache);
```
csfrancis, line 58: Maybe set these to `NULL` after freeing?

ext/statsd/statsd.c
```diff
@@ -0,0 +1,376 @@
...
+  size_t size;
+  const struct datagram_builder *builder = (struct datagram_builder *)ptr;
+  size = sizeof(struct datagram_builder);
+#ifdef NORMALIZED_TAGS_CACHE_ENABLED
+  size += st_memsize(builder->normalized_tags_cache);
+#endif
+#ifdef NORMALIZED_NAMES_CACHE_ENABLED
+  size += st_memsize(builder->normalized_names_cache);
+#endif
+  return size;
```
csfrancis, line 80: Will this include the sizes of the `VALUE` members on the builder (assuming the values can’t be embedded)?

ext/statsd/statsd.c
```diff
@@ -0,0 +1,376 @@
...
+  long chunk_len = 0;
+  get_datagram_builder_struct(self);
+
+  builder->len = builder->prefix_len;
+
+  if (NIL_P(normalized_name = normalize_name_fast_path(self, name))) {
+    normalized_name = normalized_names_cached(builder, self, name);
+  }
+  chunk_len = RSTRING_LEN(normalized_name);
+  if (builder->len + chunk_len > DATAGRAM_SIZE_MAX) goto finalize_datagram;
```
csfrancis, line 248: Should we be raising on these conditions?

ext/statsd/statsd.c
```diff
@@ -0,0 +1,376 @@
...
+  memcpy(builder->datagram + builder->len, StringValuePtr(normalized_name), chunk_len);
+  builder->len += chunk_len;
+
+  if (builder->len + 1 > DATAGRAM_SIZE_MAX) goto finalize_datagram;
```
csfrancis, line 252: You could maybe extract this into an inline function.

ext/statsd/statsd.c
methodmissing: Yes agree, use after free guard

ext/statsd/statsd.c
methodmissing: I thought about it, but then again this is telemetry specific and the default buffer is a generous buffer size of 4096 bytes per datagram.

The line is somewhere between:

* Raising on a telemetry specific path the could bubble up an unhandled exception, impacting business specific paths. I don't think that's a great default.

* If you emit datagrams that large chances are cardinality is also going to be high through tags most likely and a truncated datagram would be the least we could do.

ext/statsd/statsd.c
methodmissing: Was thinking about it too, but for flow control to convert the buffer to a Ruby String and return, the goto is local.

A static function would require length and the builder as arguments too and wouldn't make a difference to code size.

Thoughts?

ext/statsd/statsd.c
methodmissing: My understanding of object memory size reporting (as expected to be returned by `ObjectSpace.memsize_of`) is that retained memory shouldn't be included. See [Class](https://github.com/ruby/ruby/blob/12009fb3b984faf346250ee7723ecccf56622ce0/gc.c#L3937-L3954) for example 

Heap dumps follow that model too, otherwise the [dominator tree](https://github.com/csfrancis/harb/pull/6) for harb would not have been needed :smile:

[st_memsize](https://github.com/ruby/ruby/blob/0f2f07a9bb0fcbfe897e956965fcf0cfb1ffdbfa/st.c#L647-L654) is just memory consumed by the hash table internals (bins and entries)

methodmissing:
@wvanbergen objections to spawning `statsd-instrument-c` (blocked on a better name, patches welcome :smile:)? That would also introduce some other issues to think about:

* What happens with test coverage in light of the extracted extension?
* Running the main test suite with and without the C extensions (gems become a CI dependency)

ext/statsd/statsd.c
methodmissing: https://github.com/Shopify/statsd-instrument/pull/259/commits/666ff836dd4f05fdd0c92b8ab267e10f3fd91e0a

ext/statsd/statsd.c
methodmissing: Or let overflow checking be a compile time option and allow arbitrary size datagrams if disabled (effectively the current Ruby implementation's behaviour)

ext/statsd/statsd.c
csfrancis: What's the behaviour if we send a datagram that's too large? Does that depend on the statsd backend?

ext/statsd/statsd.c
csfrancis: I was thinking that this would make more sense if we were raising an exception.

ext/statsd/statsd.c
wvanbergen: Currently, we don't know. I assume they will simply not be delivered because they are being dropped by the networking stack as the UDP datagrams get too large. By design, we don't know about this - it's a fire and forget protocol.

methodmissing: What Willem said re. `fire and forget`. I'm fine with any solution other than raising an error as telemetry specific fire and forget datagrams shouldn't risk raising unhandled exceptions that could uproot business processes.

ext/statsd/statsd.c
methodmissing: The risk / regression factor in raising exceptions is that we'd violate a loose contract of "allowing any size datagrams to be built" to "raising an exception that can fan out to the whole host application"

Thoughts on a middle ground that either:

* Implements overflow guards, truncates the datagram and emit a warning with `rb_warn("StatsD datagram truncated to X bytes")`. Easy to provide a reason too, as with the logging pipeline.

* Follow the existing behavior of supporting arbitrary size datagrams: static buffer that can overflow to a heap allocated buffer OR move the buffer to the heap. Both of these aren't difficult, but creates many more branches and complexity too.

?

PR #260, opened 2020-04-13
Rubocop improvements

wvanbergen:
- Update Rubocop to latest version
- Update Shopify Ruby style guide to latest version
- Enable the few rules that previously were disabled.
- (Auto)correct all offences.

Best reviewed with whitespace diff turned on: https://github.com/Shopify/statsd-instrument/pull/260/files?w=1


rafaelfranca:
I think we should change the way this project inherit the styleguide from the URL to the gem.

you can add this to your Gemfile:

```
gem 'rubocop-shopify', require: false
```

And add to the top of your project's RuboCop configuration file:

```
inherit_gem:
  rubocop-shopify: rubocop.yml
```

Instead of the `inherit_from`. You can also delete the cache file after that.

wvanbergen:
@rafaelfranca updated

PR #265, opened 2020-05-19
Fix delegation for Ruby 2.7 compatibility

casperisfine:
`statsd-instrument` decorators are throwing warnings on Ruby 2.7:

```
DEPRECATION WARNING: /tmp/bundle/ruby/2.7.0/gems/statsd-instrument-2.9.2/lib/statsd/instrument/strict.rb:130: warning: Using the last argument as keyword parameters is deprecated; maybe ** should be added to the call
xxxxxx:103: warning: The called method `XXXXXX' is defined here
 (called from block (3 levels) in statsd_measure at /tmp/bundle/ruby/2.7.0/gems/statsd-instrument-2.9.2/lib/statsd/instrument/strict.rb:130)
```

While not very elegant, the `ruby2_keywords` modifier is the only proper way to do blind argument forwarding on both 2.6 and 2.7 without warnings.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -11,7 +11,7 @@ jobs:
       matrix:
         # Windows on macOS builds started failing, so they are disabled for noew
 
-        ruby: [2.4, 2.5, 2.6]
+        ruby: [2.4, 2.5, 2.6, 2.7]
         # platform: [windows-2019, macOS-10.14, ubuntu-18.04]
 
         # exclude:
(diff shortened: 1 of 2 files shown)


casperisfine:
Would also need to be backported to the 2.x branch.

.github/workflows/ci.yml
```diff
@@ -43,7 +43,7 @@ jobs:
       run: gem install bundler && bundle install --jobs 4 --retry 3
 
     - name: Run test suite
-      run: bin/rake test
```
wvanbergen, line 46: Does this not work with 2.7?

casperisfine: Yeah it's a bug in the bundler embeded with 2.7. The same bug that makes nested `bundle exec` fail.

wvanbergen:
I am not a big fan of the metaprogramming methods, so I don't mind using the ugly fix here. I may officially deprecate them at some point.

PR #266, opened 2020-05-20
Fix the gemspec executables prefix

casperisfine:
I just noticed this because I installed the `3.0.0` for the first time.

A while ago bundler used to register `bin/` as the executables directory, meaning the scripts in that path would be exposed as commands in the `$PATH` of users installing the gem.

The problem is that it's also where binstubs are installed, so now bundler use `exe/` instead.

So in short `statsd-instrument 3.0.0` break `rake` and `rubocop` commands if you install it.


casperisfine:
If you don't mind I'll release a 3.0.1 right away.

Issue #267, opened 2020-06-24
assertion calls with no_prefix: true will set the expectation name to nil instead of just keeping the name


SaifBenChahed:
assertion calls with no_prefix: true will set the expectation name to nil instead of just keeping the name

assert_statsd_increment('PlusMetrics.storefront.requests', no_prefix: true) do
      
end

will expect an increment with no name because of this https://github.com/Shopify/statsd-instrument/blob/fff742eabee98f1cc80f68e4bc77b13ad28cc5f8/lib/statsd/instrument/expectation.rb#L39

PR #268, opened 2020-06-29
Properly handle no_prefix when using StatsD assertions.

wvanbergen:
Fixes https://github.com/Shopify/statsd-instrument/issues/267.

### DIFF
--- a/lib/statsd/instrument/expectation.rb
+++ b/lib/statsd/instrument/expectation.rb
@@ -36,7 +36,7 @@ module StatsD
         sample_rate: nil, tags: nil, no_prefix: false, times: 1)
 
         @type = type
-        @name = client.prefix ? "#{client.prefix}.#{name}" : name unless no_prefix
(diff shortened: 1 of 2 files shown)


lib/statsd/instrument/expectation.rb
```diff
@@ -36,7 +36,7 @@ def initialize(client: StatsD.singleton_client, type:, name:, value: nil,
         sample_rate: nil, tags: nil, no_prefix: false, times: 1)
 
         @type = type
-        @name = client.prefix ? "#{client.prefix}.#{name}" : name unless no_prefix
+        @name = client.prefix && !no_prefix ? "#{client.prefix}.#{name}" : name
```
casperisfine, line 39: I couldn't comment in time because of the GH outage, but I find this condition hard to grasp.

`no_prefix || !client.prefix` would have been much more readable.

wvanbergen: Fixed on master

PR #272, opened 2020-08-13
Use published version of benchmark-ips

sambostock:
The released versions now include `save!`, as per [`History.txt`](https://github.com/evanphx/benchmark-ips/blob/5a2dc713fede28e0c08d2c9d43efad967cba6019/History.txt#L11).

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -10,6 +10,4 @@ gem 'mocha'
 gem 'yard'
 gem 'rubocop', '>= 1.0'
 gem 'rubocop-shopify', require: false
-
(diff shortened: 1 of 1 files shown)


sambostock:
I'm not sure why the benchmarks are [suddenly failing](https://github.com/Shopify/statsd-instrument/runs/982551102), but I can replicate on a [branched copy of `master`](https://github.com/Shopify/statsd-instrument/runs/982562395), so it doesn't seem related to this change.

wvanbergen (approved):
Rebased; CI is happy now.

PR #278, opened 2021-03-30
Add Ruby 3.0 to CI

casperisfine:
And enable deprecation warnings.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -11,7 +11,7 @@ jobs:
       matrix:
         # Windows on macOS builds started failing, so they are disabled for noew
 
-        ruby: [2.4, 2.5, 2.6, 2.7]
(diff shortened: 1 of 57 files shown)


casperisfine:
Argh, the recent change to rubocop-shopify cause a lot of failures. I'd ran the auto-corrector.

.github/workflows/ci.yml
```diff
@@ -11,7 +11,7 @@ jobs:
       matrix:
         # Windows on macOS builds started failing, so they are disabled for noew
 
-        ruby: [2.4, 2.5, 2.6, 2.7]
+        ruby: ['2.4', '2.5', '2.6', '2.7', '3.0']
```
wvanbergen, line 14: Feel free to drop some older versions here.

casperisfine: As long as they don't cause maintenance pain, I'm fine to keep 2.4.

PR #280, opened 2021-04-01
Implement batching of datagrams in UDPSink

casperisfine:
This is kind of an alternative to https://github.com/Shopify/statsd-instrument/pull/257 (actually we could combine both, but...)

With a batch size of one, this is a bit slower than master, however with an arbitrary batch size of 50, it's about twice faster:

```
StatsD metrics to local UDP receiver (branch: master, sha: a9fae17)
                          5.142k (± 4.2%) i/s -     26.000k in   5.067016s

Comparison:
StatsD metrics to local UDP receiver (branch: udp-batching, sha: 0f436b2):    10458.0 i/s
StatsD metrics to local UDP receiver (branch: master, sha: a9fae17):     5141.5 i/s - 2.03x  (± 0.00) slower
```

Thread safety wise, I believe `Concurrent::Array` should have us covered (it's a regular Array on MRI, so little perf overhead).

What I'm not too sure about is the flushing. I'm worried some datagrams could stay in memory for quite long. The finalizer should ensure that they are sent at some point, however they could end up delayed a lot of a process stay idle for long with some datagrams.

So maybe it could make sense to combine this with the dispatcher thread PR, to have a background thread empty the queue every seconds or something.

@wvanbergen @dylanahsmith any thoughts?

### DIFF
--- a/README.md
+++ b/README.md
@@ -42,6 +42,29 @@ The following environment variables are supported:
   overridden in a metric method call.
 - `STATSD_DEFAULT_TAGS`: A comma-separated list of tags to apply to all metrics.
   (Note: tags are not supported by all implementations.)
+- `STATSD_FLUSH_INTERVAL`: (default: `0.0`) The interval at which events are sent
+  in batch. Only applicable to the UDP configuration. If set to `0.0`, metrics
+  are sent immediately.
+
+### Forking setups
+
+If you use some kind of forking server, you must call `StatsD.after_fork` after each fork.
+
+For instance if you use Puma:
+
+```ruby
+on_worker_boot do
+  StatsD.after_fork
+end
+```
+
+Or if you use Unicorn:
+
+```ruby
+after_fork do
+  StatsD.after_fork
+end
+```
 
 ## StatsD keys
 
--- a/README.md
+++ b/README.md
@@ -46,26 +46,6 @@ The following environment variables are supported:
   in batch. Only applicable to the UDP configuration. If set to `0.0`, metrics
   are sent immediately.
 
-### Forking setups
-
-If you use some kind of forking server, you must call `StatsD.after_fork` after each fork.
-
-For instance if you use Puma:
-
-```ruby
-on_worker_boot do
-  StatsD.after_fork
-end
-```
-
-Or if you use Unicorn:
-
-```ruby
-after_fork do
-  StatsD.after_fork
-end
-```
-
 ## StatsD keys
 
 StatsD keys look like 'admin.logins.api.success'. Dots are used as namespace separators.
--- a/lib/statsd/instrument/batched_udp_sink.rb
+++ b/lib/statsd/instrument/batched_udp_sink.rb
@@ -28,17 +28,13 @@ module StatsD
         spawn_dispatcher
       end
 
-      def after_fork
-        @buffer.clear
-        spawn_dispatcher
-      end
-
       def sample?(sample_rate)
         sample_rate == 1 || rand < sample_rate
       end
 
       def <<(datagram)
         @buffer << datagram
+        spawn_dispatcher unless @dispatcher_thread&.alive?
         self
       end
 
@@ -49,9 +45,7 @@ module StatsD
       private
 
       def spawn_dispatcher
-        unless @dispatcher_thread&.alive?
-          @dispatcher_thread = Thread.new { dispatch }
-        end
+        @dispatcher_thread = Thread.new { dispatch }
       end
 
       NEWLINE = "\n".b.freeze
--- a/lib/statsd/instrument/batched_udp_sink.rb
+++ b/lib/statsd/instrument/batched_udp_sink.rb
@@ -1,0 +1,124 @@
+# frozen_string_literal: true
+
+module StatsD
+  module Instrument
+    # @note This class is part of the new Client implementation that is intended
+    #   to become the new default in the next major release of this library.
+    class BatchedUDPSink
+      DEFAULT_FLUSH_INTERVAL = 1.0
+      MAX_PACKET_SIZE = 508
+
+      def self.for_addr(addr, flush_interval: DEFAULT_FLUSH_INTERVAL)
+        host, port_as_string = addr.split(":", 2)
+        new(host, Integer(port_as_string), flush_interval: flush_interval)
+      end
+
+      attr_reader :host, :port
+
+      def initialize(host, port, flush_interval: DEFAULT_FLUSH_INTERVAL)
+        @host = host
+        @port = port
+        @mutex = Mutex.new
+        @socket = nil
+        @flush_interval = flush_interval
+
+        require "concurrent/array"
+        @buffer = Concurrent::Array.new
+        @dispatcher_thread = nil
+        spawn_dispatcher
+      end
+
+      def after_fork
+        @buffer.clear
+        spawn_dispatcher
+      end
+
+      def sample?(sample_rate)
+        sample_rate == 1 || rand < sample_rate
+      end
+
+      def <<(datagram)
+        @buffer << datagram
+        self
+      end
+
+      def shutdown
+        @dispatcher_thread&.kill
+      end
+
+      private
+
+      def spawn_dispatcher
+        unless @dispatcher_thread&.alive?
+          @dispatcher_thread = Thread.new { dispatch }
+        end
+      end
+
+      NEWLINE = "\n".b.freeze
+      def flush
+        return if @buffer.empty?
+
+        datagrams = @buffer.shift(@buffer.size)
+
+        until datagrams.empty?
+          packet = String.new(datagrams.pop, encoding: Encoding::BINARY, capacity: MAX_PACKET_SIZE)
+
+          until datagrams.empty? || packet.bytesize + datagrams.first.bytesize + 1 > MAX_PACKET_SIZE
+            packet << NEWLINE << datagrams.shift
+          end
+
+          send_packet(packet)
+        end
+      end
+
+      def dispatch
+        loop do
+          begin
+            start = Process.clock_gettime(Process::CLOCK_MONOTONIC)
+            flush
+            next_sleep_duration = @flush_interval - (Process.clock_gettime(Process::CLOCK_MONOTONIC) - start)
+            sleep(next_sleep_duration) if next_sleep_duration > 0
+          rescue => error
+            report_error(error)
+          end
+        end
+      end
+
+      def report_error(error)
+        StatsD.logger.error do
+          "[#{self.class.name}] The dispatcher thread encountered an error #{error.class}: #{error.message}"
+        end
+      end
+
+      def send_packet(packet)
+        retried = false
+        socket.send(packet, 0)
+      rescue SocketError, IOError, SystemCallError => error
+        StatsD.logger.debug do
+          "[#{self.class.name}] Resseting connection because of #{error.class}: #{error.message}"
+        end
+        invalidate_socket
+        if retried
+          StatsD.logger.warning do
+            "[#{self.class.name}] Events were dropped because of #{error.class}: #{error.message}"
+          end
+        else
+          retried = true
+          retry
+        end
+      end
+
+      def socket
+        @socket ||= begin
+          socket = UDPSocket.new
+          socket.connect(@host, @port)
+          socket
+        end
+      end
+
+      def invalidate_socket
+        @socket = nil
+      end
+    end
+  end
+end
--- a/lib/statsd/instrument/batched_udp_sink.rb
+++ b/lib/statsd/instrument/batched_udp_sink.rb
@@ -18,7 +18,6 @@ module StatsD
       def initialize(host, port, flush_interval: DEFAULT_FLUSH_INTERVAL)
         @host = host
         @port = port
-        @mutex = Mutex.new
         @socket = nil
         @flush_interval = flush_interval
 
@@ -33,8 +32,12 @@ module StatsD
       end
 
       def <<(datagram)
+        unless @dispatcher_thread&.alive?
+          @buffer.clear
+          spawn_dispatcher
+        end
+
         @buffer << datagram
-        spawn_dispatcher unless @dispatcher_thread&.alive?
         self
       end
 
@@ -89,7 +92,7 @@ module StatsD
         socket.send(packet, 0)
       rescue SocketError, IOError, SystemCallError => error
         StatsD.logger.debug do
-          "[#{self.class.name}] Resseting connection because of #{error.class}: #{error.message}"
+          "[#{self.class.name}] Resetting connection because of #{error.class}: #{error.message}"
         end
         invalidate_socket
         if retried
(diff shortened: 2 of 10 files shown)


wvanbergen:
I am not a big fan of adding a runtime dependency, but if that avoids us from having the overhead of a mutex lock for every StatsD call, that may be ok.

lib/statsd/instrument/udp_sink.rb
```diff
@@ -5,45 +5,72 @@ module Instrument
...
+      end
+
+      NEWLINE = "\n".b.freeze
+      def flush
+        return if @buffer.empty?
+
+        synchronize do
+          return if @buffer.empty?
+          messages = @buffer.shift(@buffer.size)
+          emit_datagram(messages.join(NEWLINE))
```
wvanbergen, line 51: While I know that Datadog will deal with newlines properly, I am not sure if this is the case for all StatsD server implementations. But maybe we don't have to care about that.

lib/statsd/instrument/udp_sink.rb
```diff
@@ -5,45 +5,72 @@ module Instrument
...
+
+      def initialize(host, port, batch_size: 1)
         @host = host
         @port = port
         @mutex = Mutex.new
         @socket = nil
+
+        require "concurrent/array"
+        @buffer = Concurrent::Array.new
+        @batch_size = batch_size
```
wvanbergen, line 29: Rather than batch size, I wonder whether we should instead (or in addition) measure the size of the full datagram that we are going to send. If this goes over MTU, there is a risk of the packet being dropped by the network.

The maximum safe size of a UDP packet is 508 bytes. However, in almost all cases, 1500 bytes is considered safe as well, as almost all networks are configured with an MTU around that size

So rather than building an array, does it make sense to build a string instead? This way we also prevent the `join` that would havew to iterate over the full array for every batch.

lib/statsd/instrument/udp_sink.rb
```diff
@@ -5,45 +5,72 @@ module Instrument
...
+        new(host, Integer(port_as_string), **kwargs)
       end
 
       attr_reader :host, :port
 
-      def initialize(host, port)
+      class << self
+        def finalizer(sink)
+          proc { sink.flush }
+        end
```
wvanbergen, line 18: Do we depend on the user to register this finalizer?

lib/statsd/instrument/udp_sink.rb
casperisfine: So with the current implementation of simply sending once the buffer is full, a string would be too much trouble:

```ruby
if datagram.bytesize + buffer.bytesisze > MAX_BUFFER_SIZE
  flush
end
```

However if we go with a dispatcher thread (which I think we need otherwise as you noted there's no guarantee the datagram will be send anywhere in time), then it becomes harder. What do we do if the buffer is full? We have to store the datagram somewhere, and if we append to the string the dispatcher thread will have to do complex work to cut it back in pieces.

So just quickly like this I think we should continue with the array, but just be a bit smarter than a simple `join("\n")`. We can group the datagrams by packets of 500 bytes (or whatever the config is).

lib/statsd/instrument/udp_sink.rb
casperisfine: Thanks for the info. I'm actually not that knowledgeable about statsd outside of Datadog.

I can gate all this behavior under the `datadog` implementation.

casperisfine:
Thanks for the feedback Willem. I think I'll refactor all this to include a dispatcher thread.

casperisfine:
> I think I'll refactor all this to include a dispatcher thread.

Oh, one more question for you about that. Copy pasting from https://github.com/Shopify/statsd-instrument/pull/257#issuecomment-810572574

> When forking all the thread but the main one are dead in the child process. So if we use a dispatcher thread, they need to be respawned after fork. Either automatically (e..g by checking `Process.pid` all the time, or [by decorating `Process.fork`](https://github.com/rails/rails/blob/5de2dbff16009ebf8441e259d95d4b22d068ac8c/activesupport/lib/active_support/fork_tracker.rb)) or either by requiring the users to call some method.

> I went for the later and introduced a `.after_fork` method for that purpose. But I'm open to implement "fork tracking" too. It would be easier to use, but libraries monkey patching core methods are not generally well received. 

Any opinions on this? IMO checking `Process.pid` is a bad idea as it's rather slow on Linux (always do a syscall).

wvanbergen:
> Any opinions on this? IMO checking Process.pid is a bad idea as it's rather slow on Linux (always do a syscall).

The `after_fork` approach seems reasonable to me.

lib/statsd/instrument/udp_sink.rb
wvanbergen: The original `statsd` implementation also deals with newlines correctly. I am OK with assuming this always works - we don'r really support other implementations anyway.

casperisfine:
I think this should be good to go.

I deployed it on a minor service to make sure it works alright.

Also I chose to make this batching entirely opt-in, because with the need of calling `after_fork`, I'm afraid many users would upgrade without noticing the need for calling `after_fork`. That could easily cause a situation where the dispatcher thread is dead post fork, so the datagrams would accumulate forever in the buffer. Metrics wouldn't be sent, and the app memory would grow forever.

Some quick benchmark:

```
$ benchmark/send-metrics-to-local-udp-receiver
StatsD metrics to local UDP receiver (branch: udp-batching, sha: 2575938)
                         1.351k i/100ms
StatsD metrics to local UDP receiver (branch: udp-batching, sha: 2575938)
                         13.476k (± 0.9%) i/s -     67.550k in   5.012945s

$ STATSD_FLUSH_INTERVAL=0.1 benchmark/send-metrics-to-local-udp-receiver
StatsD metrics to local UDP receiver (branch: udp-batching, sha: 2575938)
                         6.303k i/100ms
StatsD metrics to local UDP receiver (branch: udp-batching, sha: 2575938)
                         60.924k (±27.3%) i/s -    277.332k in   5.052112s
```

lib/statsd/instrument/batched_udp_sink.rb
```diff
@@ -0,0 +1,124 @@
...
+
+      def initialize(host, port, flush_interval: DEFAULT_FLUSH_INTERVAL)
+        @host = host
+        @port = port
+        @mutex = Mutex.new
+        @socket = nil
+        @flush_interval = flush_interval
+
+        require "concurrent/array"
+        @buffer = Concurrent::Array.new
```
wvanbergen, line 26: What is difference with the built-in Queue class? Less overhead?

casperisfine: No, it's just a very different usage. `Queue` allows you to block until you get an element. But it doesn't have any method allowing you to pop more than one element, so it's not suitable for batching.

`Concurrent::Array` is basically MRI's regular array, but thread safe. On MRI it's just `::Array`, for other implementations there some synchronization.

lib/statsd/instrument/batched_udp_sink.rb
```diff
@@ -0,0 +1,124 @@
...
+
+        until datagrams.empty?
+          packet = String.new(datagrams.pop, encoding: Encoding::BINARY, capacity: MAX_PACKET_SIZE)
+
+          until datagrams.empty? || packet.bytesize + datagrams.first.bytesize + 1 > MAX_PACKET_SIZE
+            packet << NEWLINE << datagrams.shift
+          end
+
+          send_packet(packet)
+        end
```
wvanbergen, line 71: What does this do if a single packet > 508 bytes?

README.md
```diff
@@ -42,6 +42,29 @@ The following environment variables are supported:
...
+on_worker_boot do
+  StatsD.after_fork
+end
+```
+
+Or if you use Unicorn:
+
+```ruby
+after_fork do
+  StatsD.after_fork
```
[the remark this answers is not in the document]
casperisfine: I removed the need for `after_fork`.

lib/statsd/instrument/batched_udp_sink.rb
```diff
@@ -0,0 +1,118 @@
...
+        sample_rate == 1 || rand < sample_rate
+      end
+
+      def <<(datagram)
+        @buffer << datagram
+        spawn_dispatcher unless @dispatcher_thread&.alive?
+        self
+      end
+
+      def shutdown
```
casperisfine, line 41: I added this method for test purpose. However if for some reason an app is creating lots of short lived sinks, I'm pretty sure the dispatcher thread will prevent them from being garbage collected, causing a memory leak and worse and thread leak.

wvanbergen (approved):
This makes sense!

- Should we add a test for a datagram that goes over the max?
- Should we add a test that makes sure we start a dispatcher after forking?

For now, let's keep the old unbuffered sink the default, and roll this out to an app to see how well it works.

casperisfine:
> * Should we add a test for a datagram that goes over the max?

Sure.

> * Should we add a test that makes sure we start a dispatcher after forking?

There's already a test that forks and then send a signal to the child for it to send a datagram from the signal handler.

casperisfine:
> * Should we add a test for a datagram that goes over the max?

Done.



> For now, let's keep the old unbuffered sink the default, and roll this out to an app to see how well it works.

Yep. Definitely the plan.

wvanbergen:
Ship!

casperisfine:
I'll merge tomorrow morning unless @dylanahsmith has any comments.

lib/statsd/instrument/batched_udp_sink.rb
```diff
@@ -0,0 +1,118 @@
...
+        host, port_as_string = addr.split(":", 2)
+        new(host, Integer(port_as_string), flush_interval: flush_interval)
+      end
+
+      attr_reader :host, :port
+
+      def initialize(host, port, flush_interval: DEFAULT_FLUSH_INTERVAL)
+        @host = host
+        @port = port
+        @mutex = Mutex.new
```
dylanahsmith, line 21: It looks like this is unused.

lib/statsd/instrument/batched_udp_sink.rb
dylanahsmith: An alternative would be to use a String buffer, which we could even use to build the datagram into directly and could be flushed when it reaches its capacity.  Even without batching, that could provide some benefits of reducing string allocations and resizing during datagram building.  Although, that datagram building refactoring is probably out of scope for this PR.

Besides, we may want to decouple the concept of a buffered sink from the concerns sink sharing (e.g. per-thread buffers are an alternative) and periodic flushing (e.g. block based buffering is an alternative)

lib/statsd/instrument/batched_udp_sink.rb
```diff
@@ -0,0 +1,118 @@
...
+      end
+
+      NEWLINE = "\n".b.freeze
+      def flush
+        return if @buffer.empty?
+
+        datagrams = @buffer.shift(@buffer.size)
+
+        until datagrams.empty?
+          packet = String.new(datagrams.pop, encoding: Encoding::BINARY, capacity: MAX_PACKET_SIZE)
```
dylanahsmith, line 58: Ideally, we would re-use the same string across packets, however, I don't see a ruby method for just setting the length to 0 (i.e. clearing the contents) without freeing the capacity.

lib/statsd/instrument/batched_udp_sink.rb
```diff
@@ -0,0 +1,118 @@
...
+        spawn_dispatcher
+      end
+
+      def sample?(sample_rate)
+        sample_rate == 1 || rand < sample_rate
+      end
+
+      def <<(datagram)
+        @buffer << datagram
+        spawn_dispatcher unless @dispatcher_thread&.alive?
```
dylanahsmith, line 37: If there is a `@dispatcher_thread` but it isn't alive, then that would be due to forking, right?  In that case, shouldn't we clear the buffer to avoid double sending the buffered datagrams?

lib/statsd/instrument/batched_udp_sink.rb
```diff
@@ -0,0 +1,118 @@
...
+        datagrams = @buffer.shift(@buffer.size)
+
+        until datagrams.empty?
+          packet = String.new(datagrams.pop, encoding: Encoding::BINARY, capacity: MAX_PACKET_SIZE)
+
+          until datagrams.empty? || packet.bytesize + datagrams.first.bytesize + 1 > MAX_PACKET_SIZE
+            packet << NEWLINE << datagrams.shift
+          end
+
+          send_packet(packet)
```
dylanahsmith, line 64: Note that it looks like linux supports sending multiple packets in a single system call with [sendmmsg](https://linux.die.net/man/2/sendmmsg), although it doesn't look ruby provides a way to use it without a native extension

lib/statsd/instrument/batched_udp_sink.rb
```diff
@@ -0,0 +1,118 @@
...
+      def socket
+        @socket ||= begin
+          socket = UDPSocket.new
+          socket.connect(@host, @port)
+          socket
+        end
+      end
+
+      def invalidate_socket
+        @socket = nil
```
dylanahsmith, line 114: If for some reason the app is creating short lived sinks, then just relying on garbage collection to close the sockets could result in file descriptors to be exhausted in between garbage collection cycles.  Did you avoid closing the socket to avoid getting an error with sending to the socket from the thread after it reads from this instance variable?

lib/statsd/instrument/batched_udp_sink.rb
casperisfine: Good catch. I forgot to add this back when I refactored the `after_fork` callback to replace it with the `alive?` check. I'll add it back.

lib/statsd/instrument/batched_udp_sink.rb
casperisfine: Yes, I went though the same idea. Unfortunately as you point it `String#clear` frees the capacity entirely.

I haven't tested `StringIO`, but from experience it tend to actually be slower than Strings.

lib/statsd/instrument/batched_udp_sink.rb
casperisfine: Yeah, would be worth to try exposing it with a small C-ext. But now that `sendmsg` is in a background thread I don't think the difference will be that big.

statsd-instrument.gemspec
```diff
@@ -21,4 +21,6 @@ Gem::Specification.new do |spec|
   spec.require_paths = ["lib"]
 
   spec.metadata['allowed_push_host'] = "https://rubygems.org"
+
+  spec.add_runtime_dependency 'concurrent-ruby'
```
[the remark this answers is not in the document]
casperisfine: Done.

casperisfine:
Ok, merging now. There are a few good ideas, but I'd rather get this out of the door and get some data in production. We can always improve it more later.

lib/statsd/instrument/batched_udp_sink.rb
dylanahsmith: I don't think short-lived periodically flushed sinks make sense outside of tests, so we could probably ignore that case for now.

Inside tests, we use `shutdown` to cleanup sinks, but we don't close the socket from that method.

We also call this method internally from `send_packet` on an error, which is a case where we should be closing the socket.

PR #281, opened 2021-04-09
Enable UDP batching by default

casperisfine:
So we deployed https://github.com/Shopify/statsd-instrument/pull/280 in production and the result are very good, and we didn't notice any issue.

Given the perf improvement, and that it no longer need an `after_fork` callback, I think we should enable it by default so that people get the perf improvement simply by updating.

@wvanbergen @dylanahsmith thoughts?

### DIFF
--- a/README.md
+++ b/README.md
@@ -42,9 +42,9 @@ The following environment variables are supported:
   overridden in a metric method call.
 - `STATSD_DEFAULT_TAGS`: A comma-separated list of tags to apply to all metrics.
   (Note: tags are not supported by all implementations.)
-- `STATSD_FLUSH_INTERVAL`: (default: `0.0`) The interval at which events are sent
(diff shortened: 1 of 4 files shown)


wvanbergen (approved):

wvanbergen:
- We should do a minor new version release after merging this.
- Does it make sense to keep the old UDP sink? Or can we replace the old implementation with the new one?

casperisfine:
> * Does it make sense to keep the old UDP sink? Or can we replace the old implementation with the new one?

I think it does still make sense yes. It would be a bit hard to have the `Batched` sink handle a `0.0` interval.

> We should do a minor new version release after merging this.

Yes, 3.1.0 is good for you?

wvanbergen:
> Yes, 3.1.0 is good for you?

:+1:

PR #284, opened 2021-04-09
Improve thread & socket handling in batched UDP sink

wvanbergen:
As discussed in #280. This PR changes the batched UDP sink so that `shutdown` will gracefully exit the dispatcher thread, and will free resources associated with the socket.

### DIFF
--- a/lib/statsd/instrument/batched_udp_sink.rb
+++ b/lib/statsd/instrument/batched_udp_sink.rb
@@ -35,6 +35,7 @@ module StatsD
         require "concurrent/array"
         @buffer = BUFFER_CLASS.new
         @dispatcher_thread = nil
+        @interrupted = false
         spawn_dispatcher
       end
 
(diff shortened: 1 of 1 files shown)


casperisfine (approved):

lib/statsd/instrument/batched_udp_sink.rb
```diff
@@ -80,11 +83,12 @@ def flush
       end
 
       def dispatch
-        loop do
+        until @interrupted
```
wvanbergen, line 86: Alternatively, we could use `pending_interrupt?`/`handle_interrupt` and `Thread.raise` for exiting the thread.

casperisfine: Nah, this is fine. I'd rather avoid `Thread.raise`.

casperisfine: Hum, actually wait. if we are interrupted we should still flush the queue.

PR #285, opened 2021-04-09
Refactor BatchedUDPSink so that it can properly be GCed

casperisfine:
By moving the thread and the socket in another object, the thread no longer keep a reference on the `BatchedUDPSink`, which means it can be GCed, and it turn it can signal the dispatcher object to shutdown.

### DIFF
--- a/lib/statsd/instrument/batched_udp_sink.rb
+++ b/lib/statsd/instrument/batched_udp_sink.rb
@@ -7,17 +7,6 @@ module StatsD
     class BatchedUDPSink
       DEFAULT_FLUSH_INTERVAL = 1.0
       MAX_PACKET_SIZE = 508
-      BUFFER_CLASS = if !::Object.const_defined?(:RUBY_ENGINE) || RUBY_ENGINE == "ruby"
-        ::Array
-      else
-        begin
-          gem("concurrent-ruby")
-        rescue Gem::MissingSpecError
-          raise Gem::MissingSpecError, "statsd-instrument depends on `concurrent-ruby` on #{RUBY_ENGINE}"
-        end
-        require "concurrent/array"
-        Concurrent::Array
-      end
 
       def self.for_addr(addr, flush_interval: DEFAULT_FLUSH_INTERVAL)
         host, port_as_string = addr.split(":", 2)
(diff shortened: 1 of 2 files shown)


lib/statsd/instrument/batched_udp_sink.rb
```diff
@@ -26,112 +15,132 @@ def self.for_addr(addr, flush_interval: DEFAULT_FLUSH_INTERVAL)
...
+          @buffer = BUFFER_CLASS.new
+          @dispatcher_thread = Thread.new { dispatch }
+        end
 
-      def spawn_dispatcher
-        @dispatcher_thread = Thread.new { dispatch }
-      end
+        def <<(datagram)
+          unless @dispatcher_thread&.alive?
+            @buffer.clear
```
wvanbergen, line 63: Not related to this PR, but why are you clearing the buffer in this case?

lib/statsd/instrument/batched_udp_sink.rb
```diff
@@ -26,112 +15,132 @@ def self.for_addr(addr, flush_interval: DEFAULT_FLUSH_INTERVAL)
...
+            send_packet(packet)
           end
         end
-      end
 
-      def report_error(error)
-        StatsD.logger.error do
-          "[#{self.class.name}] The dispatcher thread encountered an error #{error.class}: #{error.message}"
+        def dispatch
+          until @interrupted
```
wvanbergen, line 92: Did you want to change this to `begin/until` ?

casperisfine: No, I just call `flush` on exit.

casperisfine:
> We should probably try this (and the other changes we merged) out in production before cutting a release.

Yes.

casperisfine:
Hum, I'm having trouble with the GC test on CI, but somehow it passes 100% of the time locally 🤔

casperisfine:
I tested it manually, it works well, I think it's just that the object must end up on the stakc somehow. Ruby Gc is unfortunately not 100 reliable. So I removed the finalizer test as I'd rather not introduce a flaky test.

PR #293, opened 2021-08-27
Fix new custom client not being passed to expectations bug

marijacvetkovik:
While trying to use a custom StatsD client for our project (see https://github.com/Shopify/help/pull/24350) we discovered a bug. The new client was being set properly but the new custom client was not being passed into expectations so there ends up being a discrepancy with the result and the expectation. This is because there is no separate parameter for client when initializing a new expectation and the options that are being passed into `StatsD::Instrument::Expectation` do not contain the new client and so the expectation ends up defaulting to `StatsD.singleton_client`. 

Explicitly adding the new client to the options as such, and initializing it, ensures that the new client is passed up to expectations: 
```
expectation = StatsD::Instrument::Expectation.increment(metric_name, value, client: client, **options)
```

### DIFF
--- a/test/assertions_test.rb
+++ b/test/assertions_test.rb
@@ -440,4 +440,17 @@ class AssertionsTest < Minitest::Test
       StatsD.increment("incr", no_prefix: true)
     end
   end
+
+  def test_client_propagation_to_expectations
+    foo_1_metric = StatsD::Instrument::Expectation.increment("foo")
+    @test_case.assert_statsd_expectations([foo_1_metric]) do
+      StatsD.increment("foo")
+    end
+
+    client = StatsD::Instrument::Client.new(prefix: "prefix")
+    foo_2_metric = StatsD::Instrument::Expectation.increment("foo", client: client)
+    @test_case.assert_statsd_expectations([foo_2_metric]) do
+      StatsD.increment("prefix.foo")
+    end
+  end
 end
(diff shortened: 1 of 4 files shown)


test/assertions_test.rb
```diff
@@ -440,4 +440,17 @@ def test_assert_when_using_no_prefix
...
+  def test_client_propagation_to_expectations
+    foo_1_metric = StatsD::Instrument::Expectation.increment("foo")
+    @test_case.assert_statsd_expectations([foo_1_metric]) do
+      StatsD.increment("foo")
+    end
+
+    client = StatsD::Instrument::Client.new(prefix: "prefix")
+    foo_2_metric = StatsD::Instrument::Expectation.increment("foo", client: client)
+    @test_case.assert_statsd_expectations([foo_2_metric]) do
+      StatsD.increment("prefix.foo")
```
sambostock, line 453: This actually brings up an interesting question...

**Should** the expectation even be able to detect this metric, since it was made by the "default" client?

I guess maybe the answer is that the expectation doesn't actually assert which client it came from, it uses the client to detect configuration and asserts about that 🤔

test/statsd_instrumentation_test.rb
```diff
@@ -345,7 +345,7 @@ def test_statsd_count_with_injected_client
     client = StatsD::Instrument::Client.new(prefix: "prefix")
 
     ActiveMerchant::Gateway.statsd_count(:ssl_post, "ActiveMerchant.Gateway.ssl_post", client: client)
-    assert_statsd_increment("prefix.ActiveMerchant.Gateway.ssl_post", client: client) do
+    assert_statsd_increment("ActiveMerchant.Gateway.ssl_post", client: client) do
```
sambostock, line 348: I assume this changed because now we infer the `prefix` from the `client` that was passed in?

sambostock (approved):

test/statsd_instrumentation_test.rb
marijacvetkovik: Yeah it was a bug in the original test

casperisfine (approved):
LGTM

PR #303, opened 2022-06-17
TODO: Fix rubocop 1.30 compatibility

casperisfine:
I don't have the time to dig into this just yet. Help welcome.


rafaelzimmermann:
I looked into it, but I hit an incompatibility issue, looks like we will need to run Lint under 2.7 version (I guess it should be done on a different PR)

https://github.com/Shopify/statsd-instrument/pull/346

PR #304, opened 2022-06-20
Add taggig error classes in statsd_count_success

TobiasBales:
This adds a `tag_error_class` option to `StatsD.stats_count_success` to allow keeping track of failure reasons.
This feature is opt in to not blow up cardinality (with tests) and only happens in the error case (not falsey return values)

This is to allow us to move to more usage of the metaprogramming API while still keeping track of error classes.


lib/statsd/instrument.rb
```diff
@@ -110,26 +110,34 @@ def statsd_distribution(method, name, sample_rate: nil, tags: nil, no_prefix: fa
...
           end
           result
         ensure
           client ||= StatsD.singleton_client
           suffix = truthiness == false ? "failure" : "success"
           key = StatsD::Instrument.generate_metric_name(name, self, *args)
+
+          if tag_error_class && error
+            tags ||= {}
+            tags[:error] = error.class
```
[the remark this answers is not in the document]
casperisfine: Also `error_class` would be a better name I think.

lib/statsd/instrument.rb
```diff
@@ -130,6 +130,12 @@ def statsd_count_success(method, name, sample_rate: nil, tags: nil, no_prefix: f
           client ||= StatsD.singleton_client
           suffix = truthiness == false ? "failure" : "success"
           key = StatsD::Instrument.generate_metric_name(name, self, *args)
+
+          if tag_error_class && error
+            tags ||= {}
+            tags[:error_class] = error.class.name
```
casperisfine, line 136: Actually there's a small issue here. `tags:` is passed once, so we shouldn't mutate it.

You need to `dup` it first.

lib/statsd/instrument.rb
```diff
@@ -130,6 +131,12 @@ def statsd_count_success(method, name, sample_rate: nil, tags: nil, no_prefix: f
           client ||= StatsD.singleton_client
           suffix = truthiness == false ? "failure" : "success"
           key = StatsD::Instrument.generate_metric_name(name, self, *args)
+
+          if tag_error_class && error
+            tags = tags.dup || {}
+            tags[:error_class] = error.class.name
```
casperisfine, line 137: I don't know how big a deal it is, but this assumes `tags` is a `Hash`. Most of the time it's true, but it's (badly) documented as `@param [Hash<Symbol, String>, Array<String>] tags (default: nil)`

https://github.com/Shopify/statsd-instrument/blob/a13bc0499bdf9fb4be8fb7894125e8ab8aba489f/lib/statsd/instrument/client.rb#L198

But I guess it would be a pain to support all types.

lib/statsd/instrument/helpers.rb
```diff
@@ -10,6 +10,18 @@ def capture_statsd_datagrams(client: nil, &block)
 
       # For backwards compatibility
       alias_method :capture_statsd_calls, :capture_statsd_datagrams
+
+      def self.add_tag(tags, key, value)
+        tags = tags.dup || {}
+
+        if tags.is_a?(Array)
```
casperisfine, line 17: You also need to handle strings.

TobiasBales: missed that 😓

casperisfine (approved):
:shipit:

denislaliberte (approved):
the change look good to me, I left a small optional comment


it is the first time I review in this code base... is there any testing expected from reviewer

denislaliberte:
should we add the change to the changelog ?

TobiasBales:
> should we add the change to the changelog ?

probably should, I'll try and figure out how that works

lib/statsd/instrument.rb
```diff
@@ -110,18 +110,19 @@ def statsd_distribution(method, name, sample_rate: nil, tags: nil, no_prefix: fa
     # @yieldreturn [Boolean] Return true iff the return value is considered a success, false otherwise.
     # @return [void]
     # @see #statsd_count_if
```
denislaliberte, line 112: please add the new param to the yardoc comment

TobiasBales:
> the change look good to me, I left a small optional comment
> 
> it is the first time I review in this code base... is there any testing expected from reviewer

I thought about that too but what do we do in the case that it is not a hash - in this case it will just raise, we could explicitly raise or log (using puts?, not sure) and return the original tags to not break anything

Thinking about it the last option seems like the nicest, I just don't have any idea on how to log this nicely (and how to test it well)

Issue #291, opened 2021-06-09
ThreadError: can't alloc thread


casperisfine:
We're investigating a weird issue that sometime happens in job workers.

```
ThreadError: can't alloc thread
~/gems/statsd-instrument-3.1.1/lib/statsd/instrument/batched_udp_sink.rb:68:in `new'
~/gems/statsd-instrument-3.1.1/lib/statsd/instrument/batched_udp_sink.rb:68:in `<<'
~/gems/statsd-instrument-3.1.1/lib/statsd/instrument/batched_udp_sink.rb:36:in `<<'
~/gems/statsd-instrument-3.1.1/lib/statsd/instrument/client.rb:437:in `emit'
~/gems/statsd-instrument-3.1.1/lib/statsd/instrument/client.rb:317:in `ensure in latency'
```

Looking at MRI this error seem to happen when the main thread is dead: https://github.com/ruby/ruby/blob/0fb782ee38ea37fd5fe8b1f775f8ad866a82a3f0/thread.c#L1027-L1029

Which initially sounded like nonsense, but after digging more, it turns out that this error is raised from another background thread, e.g:

```ruby
Thread.new do
  loop do
    StatsD.measure ...
  end
end
```

So the MRI error makes somewhat some sense.

### Theory

I assume that somehow the main thread is indeed dead, but that a background thread is still alive and working, causing this issue.

I have to dig more to replicate this kind of scenario, until now I thought the Ruby process would abort once the main thread is dead.

tenderlove:
I'm not sure how this could happen, but I wonder if there's somehow a race condition where the main thread gets killed (or put in a "dead" state) yet another thread gets scheduled to run even though the main thread is in the process of "dying".

From the slack comments (and just thinking about this) we have the main thread, which has a running heartbeat thread, and the heartbeat thread allocates a new thread to hit statsd.  It's the allocation of the statsd thread that fails.  Is that correct? @cardy31

casperisfine:
> It's the allocation of the statsd thread that fails. Is that correct?

That is correct yes.

> I wonder if there's somehow a race condition where the main thread gets killed (or put in a "dead" state) yet another thread gets scheduled to run even though the main thread is in the process of "dying".

I have the same suspicion yes. I have a few idea how to try to replicate that. I'll try my hand at it tomorrow.

cardy31:
[...]

PR #308, opened 2022-07-06
Check wether the main thread is alive before restarting the dispatcher thread.

casperisfine:
Fix: https://github.com/Shopify/statsd-instrument/issues/291

If the main thread is already dead, we shouldn't try to respawn the
thread. Instead we can send the datagram directly.

cc @MatthewRBruce, @Mangara. Would be nice to try to craft a test case for this using your small repro script.

### DIFF
--- a/lib/statsd/instrument/batched_udp_sink.rb
+++ b/lib/statsd/instrument/batched_udp_sink.rb
@@ -64,9 +64,18 @@ module StatsD
             # If the dispatcher thread is dead, we assume it is because
             # the process was forked. So to avoid ending datagrams twice
             # we clear the buffer.
-            @buffer.clear
-            @dispatcher_thread = Thread.new { dispatch }
+            # However if the main the main thread is dead, we won't be able
+            # to spawn a new thread, so we fallback to sending our datagram directly.
+            if Thread.main.alive?
+              @buffer.clear
+              @dispatcher_thread = Thread.new { dispatch }
+            else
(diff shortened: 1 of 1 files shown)


lib/statsd/instrument/batched_udp_sink.rb
```diff
@@ -64,9 +64,18 @@ def <<(datagram)
             # If the dispatcher thread is dead, we assume it is because
             # the process was forked. So to avoid ending datagrams twice
             # we clear the buffer.
-            @buffer.clear
-            @dispatcher_thread = Thread.new { dispatch }
+            # However if the main the main thread is dead, we won't be able
+            # to spawn a new thread, so we fallback to sending our datagram directly.
+            if Thread.main.alive?
```
MatthewRBruce, line 69: I don't believe `alive?` return false when a thread is in the `aborting` state unfortunately.  We probably want this to explicitly check `if Thread.main.status != "aborting"`

MatthewRBruce: 🤦 Yeah, I totally skimmed the `Thread.main` part of that.  I just test it out and did confirm that if we avoid the raise when the main thread isn't alive every appears to work well:
```
  ensure
    puts "In ensure, Raising a new error"
    raise StandardError "In Rescue" if Thread.main.alive?
  end
  ```

MatthewRBruce (approved):

Issue #297, opened 2021-10-06
Memory leak when using BatchedUDPSink


vahe:
We noticed a memory leak on one of our Sidekiq deployments. With `rbtrace` and diffing heap dumps the indication was it was from `statsd-instrument` gem.

```
Retained STRING 755984 objects of size 221275072/228127604 (in bytes) at: /artifacts/bundle/ruby/2.7.0/gems/statsd-instrument-3.1.2/lib/statsd/instrument/datagram_builder.rb:86
```

After setting `STATSD_FLUSH_INTERVAL` to 0, the memory leak disappeared ([to use the UDPSink instead of BatchedUDPSink](https://github.com/Shopify/statsd-instrument/blob/61440b52456bd4b181f62113f25957aa7c20ce4e/lib/statsd/instrument/environment.rb#L92-L95))

Memory graphs before:
<img width=200 src="https://user-images.githubusercontent.com/1486701/136235821-d511df7a-837c-4bd0-962e-6910bb01e57c.png"/>

After:
<img width=400 src="https://user-images.githubusercontent.com/1486701/136235736-a9b7ce6f-95f7-47a7-be64-576dcf4fe1f9.png"/>

casperisfine:
Sorry, somehow I wasn't watching the repo and didn't notice this.

A couple thoughts:

  - The default 1 second interval is likely too much. We should probably reduce it to 0.1 or something.
  - We need to handle and upper size to that array. A sleep loop isn't good enough, we need to wake up the background thread if the buffer is past a certain size. I think we could do this with a `Monitor`, I need to explore that.

PR #309, opened 2022-07-07
Wake up the dispatch thread once the buffer is past a threshold

casperisfine:
Fix: https://github.com/Shopify/statsd-instrument/issues/297

The current sleep loop can lead to large memory usage if lots of events are queued rapidly. In such scenario the buffer can grow very large until the threads wake up and flush it.

To avoid this, after enqueueing a message, if the buffer reached a threshold we explicitly wake up the dispatch thread to trigger a flush.

### DIFF
--- a/lib/statsd/instrument/batched_udp_sink.rb
+++ b/lib/statsd/instrument/batched_udp_sink.rb
@@ -23,7 +23,8 @@ module StatsD
         end
       end
 
-      def initialize(host, port, flush_interval: DEFAULT_FLUSH_INTERVAL, thread_priority: DEFAULT_THREAD_PRIORITY, flush_threshold: DEFAULT_FLUSH_THRESHOLD)
+      def initialize(host, port, flush_interval: DEFAULT_FLUSH_INTERVAL, thread_priority: DEFAULT_THREAD_PRIORITY,
+        flush_threshold: DEFAULT_FLUSH_THRESHOLD)
         @host = host
         @port = port
         @dispatcher = Dispatcher.new(host, port, flush_interval, flush_threshold, thread_priority)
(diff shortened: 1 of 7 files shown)


casperisfine:
cc @vahe

vahe (dismissed):

vahe:
Oops, I approved the wrong PR. Will review this now

casperisfine:
What would be ideal would be to test this branch on the app that was suffering from increased memory usage.

casperisfine:
We tested this in production today with @vahe. Unfortunately it's not quite enough, the app still generates events much faster than they can be sent.

We saw the `@buffer` go over 1M entries. The background thread does consume events, but it appear that the problem is that if you generate too much events, then the Ruby scheduler won't let that thread take all the CPU time (because it tries to be fair).

So clearly some metrics need to be sampled, however I think we need to have a better error condition when this happens. Right now it's "not a problem" on that app, because basically sending metrics is blocking, so if you emit too many metrics you slow down the whole app rather than to grow memory.

So I think this PR is an improvement, but we also need to but some kind of max size on the buffer and either block, error out or drop events when the buffer is full.

I'll be looking at this tomorrow.

lib/statsd/instrument/batched_udp_sink.rb
```diff
@@ -6,11 +6,14 @@ module Instrument
     #   to become the new default in the next major release of this library.
     class BatchedUDPSink
       DEFAULT_FLUSH_INTERVAL = 1.0
-      MAX_PACKET_SIZE = 508
```
casperisfine, line 9: I can't remember why it was set so low. The PR introducing this value doesn't discuss it. However I took this value from other statsd clients.

PR #316, opened 2022-08-26
Refactor BatchedUDPSink to use a SizedQueue

casperisfine:
Ruby's Queue offer cheap thread switching, which improve latency.

On 3.1 and older, `push` doesn't have a timeout, so we shim it with some `Thread.pass`.


lib/statsd/instrument/batched_udp_sink.rb
```diff
@@ -61,111 +56,85 @@ def shutdown(*args)
...
+          packet = "".b
+          next_datagram = nil
+          until @buffer.closed? && @buffer.empty? && next_datagram.nil?
+            next_datagram ||= blocking ? @buffer.pop : @buffer.pop_nonblock
+            if next_datagram.nil?
+              # Either we are in non blocking mode and there was no datagram in the buffer,
+              # or either we were in blocking mode and the queue was closed.
+              # In both case there is nothing left to do.
+              break
+            end
```
etiennebarrie, line 122: I'd split this up since these are two pretty different reasons:
```suggestion
            if blocking
              next_datagram ||= @buffer.pop
              break if next_datagram.nil? # queue was closed
            else
              next_datagram ||= @buffer.pop_nonblock
              break if next_datagram.nil? # no datagram in buffer
            end
          end
```

lib/statsd/instrument/environment.rb
```diff
@@ -78,12 +78,12 @@ def statsd_default_tags
...
-      def statsd_flush_interval
-        Float(env.fetch("STATSD_FLUSH_INTERVAL", StatsD::Instrument::BatchedUDPSink::DEFAULT_FLUSH_INTERVAL))
+      def statsd_buffer_capacity
+        Integer(env.fetch("STATSD_BUFFER_CAPACITY", StatsD::Instrument::BatchedUDPSink::DEFAULT_BUFFER_CAPACITY))
       end
 
-      def statsd_buffer_capacity
-        Float(env.fetch("STATSD_BUFFER_CAPACITY", StatsD::Instrument::BatchedUDPSink::DEFAULT_BUFFER_CAPACITY))
+      def statsd_batching?
+        statsd_buffer_capacity > 0 && Float(env.fetch("STATSD_FLUSH_INTERVAL", 1.0)) > 0.0
```
etiennebarrie, line 86: `STATSD_FLUSH_INTERVAL` is deprecated but there's no deprecation message, maybe we could `warn` when we find a value here?

lib/statsd/instrument/batched_udp_sink.rb
```diff
@@ -61,111 +56,85 @@ def shutdown(*args)
...
+      class Buffer < SizedQueue
+        def push_nonblock(item)
+          push(item, true)
+        rescue ThreadError, ClosedQueueError
+          nil
         end
 
-        def initialize(host, port, flush_interval, flush_threshold, buffer_capacity, thread_priority, max_packet_size)
+        def inspect
+          "<#{self.class.name}#{object_id} capacity=#{max} size=#{size}>"
```
etiennebarrie, line 67:
```suggestion
          "<#{self.class.name}:#{object_id} capacity=#{max} size=#{size}>"
```

etiennebarrie (approved):

PR #318, opened 2022-08-31
Update README

jcmfernandes:
According to the changelog, support for `statsite` was dropped on version 3.0.0.


casperisfine:
Bypassing CLA since I don't think README tweaks need it.

casperisfine:
Thank you!

PR #319, opened 2022-09-29
Allow user to pass a lambda function to dynamic set the tags

rafaelzimmermann:
Adds the same functionality that allows dynamic metric names to tags. Now you can use lambda function that instead of a list of tags to set the metric tags.
The lambda function must accept two arguments:
  - the object the function is being called on
  - the array of arguments passed

This is useful when defining metaprogramming methods.

```
Comparison:
UDP batched (branch: dynamic_tags, sha: d650250):   264181.0 i/s
UDP batched (branch: master, sha: e69ab35):   256938.6 i/s - same-ish: difference falls within error
```

### DIFF
--- a/lib/statsd/instrument.rb
+++ b/lib/statsd/instrument.rb
@@ -34,6 +34,15 @@ module StatsD
       name.respond_to?(:call) ? name.call(callee, args).gsub("::", ".") : name.gsub("::", ".")
     end
 
+    # Generates the tags for an instrumented method.
+    # @private
+    # @return [Array[String]]
+    def self.generate_tags(tags, callee, *args)
+      return if tags.nil?
+
+      tags.respond_to?(:call) ? tags.call(callee, args) : tags
+    end
+
     # Even though this method is considered private, and is no longer used internally,
     # applications in the wild rely on it. As a result, we cannot remove this method
     # until the next major version.
@@ -61,6 +70,8 @@ module StatsD
     # @param method [Symbol] The name of the method to instrument.
     # @param name [String, #call] The name of the metric to use. You can also pass in a
     #    callable to dynamically generate a metric name
+    # @param tags [Hash, #call] The tags to be associated with the metric. You can also
+    #    pass in a callable to dynamically generate the tags key and values
     # @param metric_options (see StatsD#measure)
     # @return [void]
     def statsd_measure(method, name, sample_rate: nil, tags: nil, no_prefix: false, client: nil)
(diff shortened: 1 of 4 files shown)


lib/statsd/instrument.rb
```diff
@@ -88,6 +100,7 @@ def statsd_distribution(method, name, sample_rate: nil, tags: nil, no_prefix: fa
         define_method(method) do |*args, &block|
           client ||= StatsD.singleton_client
           key = StatsD::Instrument.generate_metric_name(name, self, *args)
+          tags = StatsD::Instrument.generate_tags(tags, self, *args)
```
casperisfine, line 103: Would be nice if that feature didn't incur an extra cost for users that don't pass a proc.

That said, there's lots of improvement that could be made to these decorators, so maybe we should just do a perf pass  afterwards.

lib/statsd/instrument.rb
```diff
@@ -61,13 +70,16 @@ def self.duration
     # @param method [Symbol] The name of the method to instrument.
     # @param name [String, #call] The name of the metric to use. You can also pass in a
     #    callable to dynamically generate a metric name
+    # @param tags [Hash, #call] The tags to be associated with the metric. You can also
+    #    pass in a callable to dinamically generate the tags key and values
```
casperisfine, line 74:
```suggestion
    #    pass in a callable to dynamically generate the tags key and values
```

lib/statsd/instrument.rb
rafaelzimmermann: I used the following class to run the benchmark with a modified version of send-metrics-to-local-udp-receiver

```
module ActiveMerchant
  class Dummy
    extend StatsD::Instrument

    def ssl_post(arg)
      "OK"
    end
+    statsd_measure(:ssl_post, "metric.name", tags: lambda { |object, args| ["key:#{args.first}"] })  #  branch
-    statsd_measure(:ssl_post, "metric.name", tags: ["key:foo"] )  # master

    def post_with_block(&block)
      block.call if block_given?
    end
  end
end
```

```
Comparison:
UDP batched (branch: dynamic_tags, sha: d650250):   264181.0 i/s
UDP batched (branch: master, sha: e69ab35):   256938.6 i/s - same-ish: difference falls within error
```

lib/statsd/instrument.rb
```diff
@@ -192,6 +206,7 @@ def statsd_count(method, name, sample_rate: nil, tags: nil, no_prefix: false, cl
         define_method(method) do |*args, &block|
```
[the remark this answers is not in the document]
casperisfine: Nevermind, it's handled by `add_to_method`

casperisfine (approved):

PR #320, opened 2022-10-04
Release 3.5.0

rafaelzimmermann:
I was not able to create a release, looks like I am missing write access.

## Version 3.5.0

- Allow user to provide a lambda function to dynamically set metric tags


casperisfine:
`statsd-instrument` is automatically released when you bump the version number.

PR #326, opened 2023-01-04
Allow user to provide expectation with times zero

rafaelzimmermann:
This allows to mix positive and negative expectations, meaning user is able to assert that a given metric is not present while checking for the presence of others.

```
foo_1_metric = StatsD::Instrument::Expectation.increment("counter", times: 2, tags: ["foo:1"])
foo_2_metric = StatsD::Instrument::Expectation.increment("counter", times: 0, tags: ["foo:2"])
assert_statsd_expectations([foo_1_metric, foo_2_metric]) do
  ...
end
```

Without this change, if user passes `times: 0` as expectation the test does not fail and the metric is not checked to make sure it was not called, given an impression to the user that the feature is available.

### DIFF
--- a/lib/statsd/instrument/assertions.rb
+++ b/lib/statsd/instrument/assertions.rb
@@ -165,6 +165,8 @@ module StatsD
           filtered_datagrams = datagrams.select { |m| m.type == expectation.type && m.name == expectation.name }
 
           if filtered_datagrams.empty?
+            next if expectation_times == 0
+
             flunk("No StatsD calls for metric #{expectation.name} of type #{expectation.type} were made.")
           end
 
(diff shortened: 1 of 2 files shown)


lib/statsd/instrument/assertions.rb
```diff
@@ -193,7 +195,7 @@ def assert_statsd_expectations(expectations, datagrams: nil, client: nil, &block
         end
         expectations -= matched_expectations
 
-        unless expectations.empty?
+        unless expectations.select { |m| m.times != 0 }.empty?
```
qubyte, line 198: This feels like a double negative with the filter added... how about something like:

```suggestion
        if expectations.any? { |m| m.times != 0 }
```

rafaelzimmermann: Good point! I integrated your suggestion

leandromoreira (approved):
nice 🚀

PR #329, opened 2023-02-08
Adds support for prefixes to `assert_no_statsd_calls`

MiniLuke:
While using `assert_no_statsd_calls`, I found it doesn't currently support prefixes which can lead to false successes -- so this PR adds that support.

The current assertions generally expect an un-prefixed metric similar to how you'd emit the metric originally:
```
client = StatsD::Instrument::Client.new(prefix: "prefix")
assert_statsd_increment("counter", client: client) do
  client.increment("counter")
end
# Success because `prefix.counter` is emitted!
```

With `assert_no_statsd_calls`, this is actually succeeding currently:
```
client = StatsD::Instrument::Client.new(prefix: "prefix")
assert_no_statsd_calls("counter", client: client) do
  client.increment("counter")
end
# Success even though `prefix.counter` was emitted?!
```

This is because the metric names passed into `assert_no_statsd_calls` are never processed to append prefixes to them if a prefix was available, but are just used raw:
https://github.com/Shopify/statsd-instrument/blob/d5bfc145a06beaba2467170f66782a77057e8bc3/lib/statsd/instrument/assertions.rb#L64

For the positive assertions, they have the prefix appended in the `Expectation` class:
https://github.com/Shopify/statsd-instrument/blob/d5bfc145a06beaba2467170f66782a77057e8bc3/lib/statsd/instrument/expectation.rb#L42

This is especially insidious in an `assert_no_...` method when you're looking for the absence of something. Since the names aren't processed identically and the comparison finds no matches, it will always falsely succeed if a prefix is defined.

### DIFF
--- a/lib/statsd/instrument/assertions.rb
+++ b/lib/statsd/instrument/assertions.rb
@@ -73,7 +73,14 @@ module StatsD
         formatted_metrics = if no_prefix
           metric_names
         else
-          metric_names.map { |metric| StatsD::Instrument::Expectation.prefix_metric(metric, client: client) }
+          metric_names.map do |metric|
+            if StatsD::Instrument::Helpers.prefixed_metric?(metric, client: client)
+              warn("`#{__method__}` will prefix metrics by default. `#{metric}` skipped due to existing prefix.")
+              metric
+            else
+              StatsD::Instrument::Helpers.prefix_metric(metric, client: client)
+            end
+          end
         end
 
         datagrams.select! { |metric| formatted_metrics.include?(metric.name) } unless formatted_metrics.empty?
--- a/lib/statsd/instrument/assertions.rb
+++ b/lib/statsd/instrument/assertions.rb
@@ -49,6 +49,13 @@ module StatsD
       # @param [Array<String>] metric_names (default: []) The metric names that are not
       #   allowed to happen inside the block. If this is set to `[]`, the assertion
       #   will fail if any metric occurs.
+      # @param [Array<StatsD::Instrument::Datagram>] datagrams (default: nil) The datagrams
+      #   to be inspected for metric emission.
+      # @param [StatsD::Instrument::Client] client (default: nil) The client to be used
+      #   for fetching datagrams (if not provided) and metric prefix. If not provided, the
+      #   singleton client will attempt to be used.
+      # @param [Boolean] no_prefix (default: false) A directive to indicate if the client's
+      #   prefix should be prepended to the metric names.
       # @yield A block in which the specified metric should not occur. This block
       #   should not raise any exceptions.
       # @return [void]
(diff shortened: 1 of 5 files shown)


lib/statsd/instrument/assertions.rb
```diff
@@ -49,19 +49,34 @@ module Assertions
       # @param [Array<String>] metric_names (default: []) The metric names that are not
       #   allowed to happen inside the block. If this is set to `[]`, the assertion
       #   will fail if any metric occurs.
+      # @param [Array<StatsD::Instrument::Datagram>] datagrams (default: nil) The datagrams
+      #   to be inspected for metric emission.
+      # @param [StatsD::Instrument::Client] client (default: nil) The client to be used
+      #   for fetching datagrams (if not provided) and metric prefix. If not provided, the
+      #   singleton client will attempt to be used.
+      # @param [Boolean] no_prefix (default: false) A directive to indicate if the client's
```
casperisfine, line 57: I think `prefix: false` would be much more idomatic.

MiniLuke: I do think the positive case would read better, but all the existing assertions that take parameters for this take it in the form of `no_prefix` and I'd rather keep consistency.

rwstauner (commented):
what effect will this have on existing repos?

- for "assert no x" that is currently passing because there is no prefix check:
    - if the metric is present it will fail ✅ 
    - if the metric is not present it will succeed ✅ 
- for "assert no prefix.x" that is currently passing because that's the workaround:
    - the code will now look for "prefix.prefix.x" and falsely report no failures ❌ 

is that correct?
are there other scenarios to consider?
maybe we should check if metric_names already start with the prefix and emit a warning?

MiniLuke:
That makes sense to me to make it more resilient to test that had already accounted for this and emit a warning instead.

lib/statsd/instrument/expectation.rb
```diff
@@ -30,16 +30,20 @@ def distribution(name, value = nil, **options)
         def histogram(name, value = nil, **options)
           new(type: :h, name: name, value: value, **options)
         end
+
+        def prefix_metric(metric_name, client: nil)
```
rwstauner, line 34: is this used?  it looks like you moved it to Helpers.

test/assertions_test.rb
```diff
@@ -74,6 +74,66 @@ def test_assert_no_statsd_calls
...
+      @test_case.assert_no_statsd_calls("counter", client: client, no_prefix: true) do
+        client.increment("counter", no_prefix: true)
+      end
+    end
+    assert_equal(assertion.message, "No StatsD calls for metric counter expected.")
+
+    StatsD.singleton_client = client
+
+    @test_case.assert_no_statsd_calls("counter") do
+      client.increment("other")
```
rwstauner, line 119: should this one be
```suggestion
      StatsD.increment("other")
```

rwstauner (changes requested):
just a few leftovers

Issue #300, opened 2022-04-09
trigger_statsd_increment does not support multiple calls with different tags


Noezor:
👋 , thansk for the great library

# Issue

My use-case is that I have a method which issues a bunch of `statsD` calls to the same metric with different tags. I then want to verify that this metric has correctly received the expected amount of calls, each with their respective tags. Here is a test I would expect to pass: 

```ruby
require 'statsd-instrument'

RSpec.configure do |config|
  config.include StatsD::Instrument::Matchers
end

RSpec.describe 'Matchers' do
  context 'trigger_statsd_increment' do
    it 'will pass if every statsD call matches its calls' do
      expect do
        StatsD.increment('counter', tags: ['variation:a'])
        StatsD.increment('counter', tags: ['variation:b'])
      end.to trigger_statsd_increment('counter', times: 2, tags: [['variation:a'], ['variation:b']])
    end
  end
end
```

However, today this test fails. 

https://github.com/Shopify/statsd-instrument/blob/5518d8de15c838b24a987871c7b54ccb90f3a7c3/lib/statsd/instrument/matchers.rb#L59
will test that for each call to `increment`, all the tags expected are present. We are not allowed to do different tags per calls. 

Thus, likewise 
```ruby
      end.to trigger_statsd_increment('counter', times: 2, tags: ['variation:a', 'variation:b'])
```
would not work either

Rewriting the test as 
```ruby
RSpec.describe 'Matchers' do
  context 'trigger_statsd_increment' do
    it 'will pass if every statsD call matches its calls' do
      expect do
        StatsD.increment('counter', tags: ['variation:a'])
        StatsD.increment('counter', tags: ['variation:b'])
      end.to trigger_statsd_increment('counter', times: 1, tags: ["variation:a"]).and trigger_statsd_increment('counter', times: 1, tags: ["variation:b"])
    end
  end
end
```
Does not pass either and fails with 
```
The numbers of StatsD calls for metric counter was unexpected. Expected 1, got 2
```
Indeed
https://github.com/Shopify/statsd-instrument/blob/5518d8de15c838b24a987871c7b54ccb90f3a7c3/lib/statsd/instrument/matchers.rb#L45-L53 
will first test for the metric name and return the error above as the error has not been matched enough times. The tags are not looked at for this to fail.

# Propositions

##  A: support `trigger_statsd_increment('counter', times: 2, tags: [['variation:a'], ['variation:b']])`

[...]

PR #331, opened 2023-02-28
Allow user to assert different tags using RSpec

rafaelzimmermann:
Previously the matcher filtered out metrics based on type and name to test against expectation. So, when the user tried to assert different different tags, the matcher would fail.
With this change, the matcher will filter out metrics also based on tags too, allowing the user to use RSpec `and` expectation.

For instance this test would fail before:

```
RSpec.describe 'Matchers' do
  context 'trigger_statsd_increment' do
    it 'will pass if every statsD call matches its calls' do
      expect do
        StatsD.increment('counter', tags: ['variation:a'])
        StatsD.increment('counter', tags: ['variation:b'])
      end.to trigger_statsd_increment('counter', times: 1, tags: ["variation:a"]).and trigger_statsd_increment('counter', times: 1, tags: ["variation:b"])
    end
  end
end
```

fixes: https://github.com/Shopify/statsd-instrument/issues/300

### DIFF
--- a/README.md
+++ b/README.md
@@ -355,6 +355,13 @@ RSpec.describe 'Matchers' do
     it 'will pass if there is no matching StatsD call on negative expectation' do
       expect { StatsD.increment('other_counter') }.not_to trigger_statsd_increment('counter')
     end
+
+    it 'will pass if every statsD call matches its call tag variations' do
+      expect do
+        StatsD.increment('counter', tags: ['variation:a'])
+        StatsD.increment('counter', tags: ['variation:b'])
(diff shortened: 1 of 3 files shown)


casperisfine (approved):

ChanChar:
@rafaelzimmermann this may break existing tests that didn't specify tags before. 

```
def test_statsd_increment_compound_without_explicit_tags_using_and_matched
  matcher_1 = StatsD::Instrument::Matchers::Increment.new(:c, "counter", times: 1)
  matcher_2 = StatsD::Instrument::Matchers::Increment.new(:c, "counter", times: 1)

  assert(matcher_1.and(matcher_2).matches?(lambda {
    StatsD.increment("counter", tags: ["a"])
    StatsD.increment("counter", tags: ["b"])
  }))
end
```

PR #343, opened 2023-05-09
Stop variable shadowing from breaking dynamic tags

avdgaag:
## Summary

Fix dynamic tags being evaluated only once.

## Description

Re-assigning the `tags` variable defined outside a block inside the block overrides that value for subsequent calls. That caused dynamic tags to be evaluated once and then be fixed to that first value — rather than being dynamically evaluated on every invocation. By using a unique variable name, this is prevented from happening.

## Example

See the added tests; they fail without the change in this PR.


casperisfine:
@avdgaag thanks for the fix. However, sorry but could you sign the CLA thing? :/

PR #345, opened 2023-05-29
Note on repos that change logging in dev [ci skip]

larry-reid-shopify:
GIve people a hint that individual repos may change how metrics are emitted in development or test.


toneymathews (approved):

casperisfine:
statsd-instrument is an open source gem, referring to details of our private repos is not a good idea. I'll revert this.

PR #353, opened 2023-08-02
Update CONTRIBUTING.md

Mangara:
There's no need to run `bundle exec rake release`, the deployment pipeline does that for us.

I'm not sure if I should link to it, I think it's internal-only?


casperisfine (approved):
Yeah, no link please.

PR #366, opened 2024-06-06
wip: testing aggregation

pedro-stanaka:
Signed-off-by: Pedro Tanaka <pedro.tanaka@shopify.com>

### DIFF
--- a/lib/statsd/instrument/counter_aggregator.rb
+++ b/lib/statsd/instrument/counter_aggregator.rb
@@ -1,0 +1,42 @@
+# frozen_string_literal: true
+
+class CounterAggregator
+  def initialize(sink)
+    @sink = sink
+    @counters = {}
+  end
+
+  def increment(name, value = 1, sample_rate: 1.0, tags: nil)
+    key = packet_key(name, tags)
+    if @counters.has_key?(key)
+      if sample_rate < 1.0
+        value = (value.to_f / sample_rate).round.to_i
+      end
+      @counters[key][:value] += value
+    else
+      @counters[key] = {
+        name: name,
+        value: value,
+        tags: tags,
+      }
+    end
+    @counters[key] ||= {
+      name: name,
+      tags: tags,
+    }
+  end
+
+  def flush
+
+  end
+
+  private
+  def packet_key(name, tags)
+    if tags.is_a?(Hash)
+      "#{name}#{tags.sort_by { |k, v| k.to_s }.to_s}"
+    else
+      tags.sort!
+      "#{name}#{tags.join('')}"
+    end
+  end
+end
--- a/Gemfile
+++ b/Gemfile
@@ -8,6 +8,7 @@ gem "minitest"
 gem "rspec"
 gem "mocha"
 gem "yard"
+gem "racc"
 gem "rubocop", ">= 1.0"
 gem "rubocop-shopify", require: false
 gem "benchmark-ips"
--- a/test/counter_aggregator_test.rb
+++ b/test/counter_aggregator_test.rb
@@ -1,0 +1,24 @@
+# frozen_string_literal: true
+
+require "test_helper"
+
+class CounterAggregatorTest < Minitest::Test
+  def setup
+    @sink = CaptureSink.new(NullSink.new)
+    @subject = CounterAggregator.new(@sink)
+  end
+
+  def teardown
+    @sink.clear
+  end
+
+  def test_increment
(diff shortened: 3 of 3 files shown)


casperisfine:
Are you sure your profiles are correct? Because `CounterAggregator` is nowhere to be seen in the profile which is suspicious.

As for the feature more generally, I'm not very enthusiastic about this, it complexifies the code quite a bit for just counters, and in process based, so there isn't that much to aggregate.

IMO it would be much better to explore a `dogstatsd` like solution that can aggregate metrics from the whole host: https://docs.datadoghq.com/developers/dogstatsd/data_aggregation/. The datadog agent is open source, so we might be able to use it with little to no modifications: https://github.com/DataDog/datadog-agent/tree/main

pedro-stanaka:
> I'm talking about a node local relay. That's how dogstatsd is meant to be deployed.

We already talked in private, but just to leave documented here. We are already aggregating at the node level (with a DaemonSet), but since pod communicate with Daemonset via IP (not loopback) we still get aggressive network usage and also CPU usage on the Daemonset to handle all those samples.

test/client_test.rb
```diff
@@ -63,6 +63,36 @@ def test_client_from_env_with_overrides
...
+    assert_equal(["shard:1", "env:production"], client.default_tags)
+    assert_equal(StatsD::Instrument::DogStatsDDatagramBuilder, client.datagram_builder_class)
+
+    client.increment("foo", 1, sample_rate: 0.5, tags: { foo: "bar" })
+    client.increment("foo", 1, sample_rate: 0.5, tags: { foo: "bar" })
+    client.force_flush
+
+    datagram = client.sink.datagrams.first
+    assert_equal("bar.foo", datagram.name)
+    assert_equal(4, datagram.value)
```
fbogsany, line 93: This is incorrect. The value should be 2, because we incremented twice with a value of 1. The `Client` should apply sampling before aggregation, or ignore the sample rate and simply aggregate the value as supplied. Sampling before aggregation is non-deterministic, though, so you'd have to mock `rand` to have a deterministic test in that case.

lib/statsd/instrument/counter_aggregator.rb
```diff
@@ -0,0 +1,116 @@
...
+      # @param sample_rate [Float] The sample rate to use for sampling.
+      # @param tags [Hash{String, Symbol => String},Array<String>] The tags to attach to the counter.
+      # @param no_prefix [Boolean] If true, the metric will not be prefixed.
+      # @return [void]
+      def increment(name, value = 1, sample_rate: 1.0, tags: [], no_prefix: false)
+        tags = tags_sorted(tags)
+        key = packet_key(name, tags, no_prefix)
+        if sample_rate < 1.0
+          value = (value.to_f / sample_rate).round.to_i
+        end
```
fbogsany, line 65: This is only valid if we apply sampling in the `Client` before calling this method. Otherwise we should leave the `value` unchanged and ignore the `sample_rate`.

lib/statsd/instrument/client.rb
```diff
@@ -201,6 +207,12 @@ def initialize(
       # @return [void]
       def increment(name, value = 1, sample_rate: nil, tags: nil, no_prefix: false)
         sample_rate ||= @default_sample_rate
+
+        if @enable_aggregation
+          @counter_agg.increment(name, value, sample_rate: sample_rate, tags: tags, no_prefix: no_prefix)
+          return StatsD::Instrument::VOID
+        end
```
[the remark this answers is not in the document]
pedro-stanaka: I decided to do as the datadog folks are doing in [DataDog/datadog-go](https://github.com/DataDog/datadog-go/blob/1139efe797e4760daf3d30a60e3d30afc1687df8/statsd/statsd.go#L713-L722) and just ignore the sampling rate altogether. Fixed tests accordingly.

ahayworth:
> This benchmark is to be taken with a huge grain of salt in this context, because it wasn't designed for aggregation. The only counter in the benchmark has no tag at all, so it aggregates extremely well. In a real world scenario aggregation impose a noticeable overhead both in compute and memory.

That's a good point; it's the ideal case for aggregation. Perhaps a benchmark with more realistic data might be useful here - and perhaps it would be useful more generally for the library?

pedro-stanaka:
Closing and marking as superseded by #371, where we have a more complete approach using value packing for distributions/timings and also aggregate gauges.

PR #368, opened 2024-06-14
Benchmarks: adding benchmarks with more series and compare against other libs

pedro-stanaka:
## Summary

Our benchmarks currently focus solely in serialization, we would benefit from having some tests focused also on scenarios where we have more cardinality. In this PR I am introducing a SERIES_COUNT parameter to inject different label combinations on the metrics we produce and also I am including the time to close/shutdown the client on the processing time, since we might clear buffers and finish actually pushing metrics.

Apart from that, I am also comparing our implementation against the official one from Datadog.


The new benchmark results on my machine (Apple M3 Pro (11) @ 4.06 GHz) are:

```
▶ time ITERATIONS=10 SERIES_COUNT=4000 THREAD_COUNT=10 ./benchmark/local-udp-throughput                                                                                                                                                                                                               🕙 10:12:44 
===== UDP sync throughput (10 threads) =====
events: 141,326.1/s
===== UDP batched throughput (10 threads) =====
events: 306,235.9/s
===== Datadog Client - single thread throughput (10 threads) =====
events: 299,073.5/s
===== Datadog Client - multi-thread throughput (10 threads) =====
events: 1,014,260.5/s
ITERATIONS=10 SERIES_COUNT=4000 THREAD_COUNT=10   21.93s user 24.03s system 129% cpu 35.593 total
```

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -11,7 +11,4 @@ gem "yard"
 gem "rubocop", ">= 1.0"
 gem "rubocop-shopify", require: false
 gem "benchmark-ips"
-
-group :development do
-  gem "dogstatsd-ruby", "~> 5.0", require: false
-end
+gem "dogstatsd-ruby", "~> 5.0", require: false
--- a/Gemfile
+++ b/Gemfile
@@ -11,3 +11,7 @@ gem "yard"
 gem "rubocop", ">= 1.0"
 gem "rubocop-shopify", require: false
 gem "benchmark-ips"
+
+group :development do
+  gem "dogstatsd-ruby", "~> 5.0", require: false
+end
--- a/benchmark/local-udp-throughput
+++ b/benchmark/local-udp-throughput
@@ -72,7 +72,7 @@ EVENTS_PER_ITERATION = 6
 ITERATIONS = Integer(ENV.fetch("ITERATIONS", 10_000))
 SERIES_COUNT = Integer(ENV.fetch("SERIES_COUNT", 0))
 
-def benchmark_implementation(name, env = {})
+def benchmark_implementation(name, env = {}, datadog_client = false)
   intermediate_results_filename = "#{Dir.tmpdir}/statsd-instrument-benchmarks/"
   log_filename = "#{Dir.tmpdir}/statsd-instrument-benchmarks/#{File.basename($PROGRAM_NAME)}-#{name}.log"
   FileUtils.mkdir_p(File.dirname(intermediate_results_filename))
@@ -91,7 +91,7 @@ def benchmark_implementation(name, env = {})
     "STATSD_ENV" => "production",
   ).merge(env)).client
 
-  if name.start_with?("Datadog Client")
+  if datadog_client
     statsd = Datadog::Statsd.new(receiver.addr[2], receiver.addr[1], **env)
     udp_client = DatadogShim.new(statsd)
   end
@@ -111,7 +111,7 @@ def benchmark_implementation(name, env = {})
   end
   start = Process.clock_gettime(Process::CLOCK_MONOTONIC)
   threads.each(&:join)
-  if name.start_with?("Datadog Client")
+  if datadog_client
     udp_client.close
   end
   receiver.close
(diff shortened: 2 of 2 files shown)


ahayworth:
Sadly, I think the dogstatsd-ruby implementation is achieving such high rates by way of dropping things on the floor. [This monkey patch](https://github.com/Shopify/statsd-instrument/compare/feat/benchmarks...ahayworth/benchmark-sadness) shows how we can grab the internal dogstatsd-ruby telemetry and print it out while a benchmark is running... and then if I do some local testing I can see some intriguing results:

```
zsh ❯ SERIES_COUNT=100 ./local-udp-throughput
===== Datadog Client - multi-thread throughput (5 threads) =====
bytes_sent: 9817
bytes_dropped: 70141833
bytes_dropped_queue: 70141833
bytes_dropped_writer: 0
packets_sent: 10
packets_dropped: 3940561
packets_dropped_queue: 3940561
packets_dropped_writer: 0
----------
bytes_sent: 9817
bytes_dropped: 2716252
bytes_dropped_queue: 4113941
bytes_dropped_writer: 0
packets_sent: 4
packets_dropped: 466903
packets_dropped_queue: 546131
packets_dropped_writer: 0
----------
bytes_sent: 11228
bytes_dropped: 5636044
bytes_dropped_queue: 8471648
bytes_dropped_writer: 0
packets_sent: 5
packets_dropped: 951310
packets_dropped_queue: 1109485
packets_dropped_writer: 0
----------
```

I'm not immediately sure what to make of it, because those numbers seem extraordinarily high. Our benchmark might actually be causing some really suboptimal behavior in the dogstatsd-ruby client, perhaps.

benchmark/local-udp-throughput
```diff
@@ -16,13 +55,28 @@ def send_metrics(client)
...
 EVENTS_PER_ITERATION = 6
-ITERATIONS = 50_000
+ITERATIONS = Integer(ENV.fetch("ITERATIONS", 10_000))
+SERIES_COUNT = Integer(ENV.fetch("SERIES_COUNT", 0))
+
 def benchmark_implementation(name, env = {})
   intermediate_results_filename = "#{Dir.tmpdir}/statsd-instrument-benchmarks/"
   log_filename = "#{Dir.tmpdir}/statsd-instrument-benchmarks/#{File.basename($PROGRAM_NAME)}-#{name}.log"
   FileUtils.mkdir_p(File.dirname(intermediate_results_filename))
+  FileUtils.mkdir_p(File.dirname(log_filename))
```
casperisfine, line 79: I added this otherwise it would crash on my machine.

benchmark/local-udp-throughput
```diff
@@ -37,23 +91,39 @@ def benchmark_implementation(name, env = {})
...
+        end
       end
     end
   end
   start = Process.clock_gettime(Process::CLOCK_MONOTONIC)
   threads.each(&:join)
-  duration = Process.clock_gettime(Process::CLOCK_MONOTONIC) - start
-  events_sent = THREAD_COUNT * EVENTS_PER_ITERATION * ITERATIONS
-  puts "events: #{(events_sent / duration).round(1)}/s"
+  if name.start_with?("Datadog Client")
```
casperisfine, line 114: Would be better to pass as an argument.

Gemfile
```diff
@@ -11,3 +11,7 @@ gem "yard"
 gem "rubocop", ">= 1.0"
 gem "rubocop-shopify", require: false
 gem "benchmark-ips"
+
+group :development do
```
casperisfine, line 15: No need for that. In gems the entirety of the Gemfile is for development.

casperisfine (approved):

benchmark/local-udp-throughput
```diff
@@ -16,13 +55,28 @@ def send_metrics(client)
   client.service_check("StatsD.service_check", "ok")
 end
 
+def send_metrics_high_cardinality(client)
```
casperisfine, line 58: What would be interesting would be to just have an argument or env var to define the cardinality / duplication count.

pedro-stanaka: This what `SERIES_COUNT` is controlling, how many different series we are sending. Is that what you meant?

For example, for SERIES_COUNT=2

we are going to send a counter with `#series:1` and another with `series:0` for each `ITERATIONS`.

casperisfine: Yeah, I don't care so much how it's done, All I'm saying is that we should be careful to make comparisons with a fixed number of emitted metrics. What we should compare is performance on a given duplicate ratio.

PR #371, opened 2024-07-04
Using aggregation and value packing to reduce network chat

pedro-stanaka:
This is an extension of the work in #366, started as experimentation, but showed great potential in reducing not only network chat, but also CPU usage in the Ruby applications we use in production.


lib/statsd/instrument/aggregator.rb
```diff
@@ -0,0 +1,246 @@
...
+            sleep(@flush_interval)
+            thread_healthcheck
+            flush
+          end
+        end
+
+        ObjectSpace.define_finalizer(
+          self,
+          self.class.finalize(@aggregation_state, @sink, @datagram_builders, @datagram_builder_class),
+        )
```
casperisfine, line 77: I'd heavily advise against using a finalizer, especially one that does IOs.

casperisfine: `BatchedDUPSink` has one, but honestly I think it was a mistake.

pedro-stanaka: I see, I got the idea from our Sink code, but do you know of a better option in this case? I am not so versed with this part of Ruby if I am being honest.

lib/statsd/instrument/aggregator.rb
```diff
@@ -0,0 +1,246 @@
...
+        mutex.synchronize do
+          unless aggregation_state.key?(key)
+            aggregation_state[key] = {
+              type: GAUGE,
+              name: name,
+              value: value,
+              tags: tags,
+              no_prefix: no_prefix,
+            }
+          end
```
casperisfine, line 153:
```suggestion
          aggregation_state[key] ||= {
            type: GAUGE,
            name: name,
            value: value,
            tags: tags,
            no_prefix: no_prefix,
          }
        end
```

lib/statsd/instrument/aggregator.rb
```diff
@@ -0,0 +1,246 @@
...
+          # keep the last value
+          aggregation_state[key][:value] = value
+        end
+      end
+
+      def flush
+        mutex.synchronize(&method(:do_flush))
```
casperisfine, line 160:
```suggestion
        mutex.synchronize { do_flush }
```

lib/statsd/instrument/aggregator.rb
```diff
@@ -0,0 +1,246 @@
...
+          tags = sorted.map! { |k, v| "#{k}:#{v}" }
+        else
+          tags.sort!
+        end
+        tags
+      end
+
+      def packet_key(name, tags = [], no_prefix = false, type = COUNT)
+        "#{name}#{tags.join}#{no_prefix}#{type}".b
+      end
```
casperisfine, line 213: If you're going to create a cache key, might as well do it in a way that allow you to re-use its element to generate the final packet.

e.g. tags should be formatted so that you can just interpolate them in the final packet, etc.

lib/statsd/instrument/aggregator.rb
```diff
@@ -0,0 +1,246 @@
...
+          return false unless Thread.main.alive?
+
+          if @pid != Process.pid
+            StatsD.logger.info { "[#{self.class.name}] Restarting the flush thread after fork" }
+            @pid = Process.pid
+            @aggregation_state.clear
+          else
+            StatsD.logger.info { "[#{self.class.name}] Restarting the flush thread" }
+          end
+          @flush_thread = Thread.new do
```
casperisfine, line 233: This is called from unsynchonized places. All the `unless thread_healthcheck` may cause a race condition.

lib/statsd/instrument/aggregator.rb
casperisfine: Also it would be best to use a proper object rather than a Hash. Hashes aren't great for performance and compactness.

ahayworth: @casperisfine would you recommend something like a `Struct` (or maybe the new `Data`, if we are to live on the edge)? Or is defining a regular class the way to go?

(I'd also love to know where I could read more about a `Hash` not being great for performance, I had always assumed it would be heavily optimized given how prevalent it is in ruby code...)

lib/statsd/instrument/aggregator.rb
ahayworth: I presume the better option would just be to let the aggregated lines drop on the floor, and not bother trying to export them. In the common case of a `fork` we can make sure that we `flush` appropriately, and otherwise just let the data go away.

I'd be interested to know more about why finalizers are not something you'd recommend as well; it's hard to find good references about this stuff. ❤️

lib/statsd/instrument/aggregator.rb
```diff
@@ -0,0 +1,246 @@
...
+                  create_datagram_builder(datagram_builder_class, no_prefix: counter[:no_prefix])
+              end
+              sink << datagram_builders[counter[:no_prefix]].c(
+                counter[:name],
+                counter[:value],
+                CONST_SAMPLE_RATE,
+                counter[:tags],
+              )
+            end
+            aggregation_state.clear
```
ahayworth, line 28:
```suggestion
            aggregation_state.each do |_key, counter|
              datagram_builders[counter[:no_prefix]] ||= datagram_builder_class.new(
                prefix: no_prefix ? nil : @metric_prefix,
                default_tags: @default_tags,
              )
              sink << datagram_builders[counter[:no_prefix]].c(
                counter[:name],
                counter[:value],
                CONST_SAMPLE_RATE,
                counter[:tags],
              )
            end
```

You could just inline this one location where we call the `create_datagram_builder` method (unless you were trying to silence rubocop, or something).

I'd also drop the `aggregation_state.clear` since this is running in the context of a finalizer - and everything is (theoretically) about to be garbage collected anyways.

lib/statsd/instrument/aggregator.rb
casperisfine: A good old PORO.

> a Hash not being great for performance, I had always assumed it would be heavily optimized given how prevalent it is in ruby code...)

Hash is well optimized, but it's still a Hash lookup, whereas instance variables and method calls have inline caches that allow not to have to do a hash lookup in most cases. 

You can read more at https://railsatscale.com/2023-10-24-memoization-pattern-and-object-shapes/

lib/statsd/instrument/aggregator.rb
casperisfine: Because finalizers can't be run concurrently, so doing potentially long things in a finalizer may stall execution for a long time.

The reason I did it in `UDPBatchSink` was to not lose events on exit, but I think we need to start accepting that StatsD events are best effort.

lib/statsd/instrument/aggregator.rb
ahayworth: Worth remembering @pedro-stanaka that anything in ruby can be a hash key, so for example:

```ruby
irb(main):001:0> cache = {}
=> {}
irb(main):002:0> tags = ["foo:bar", "a:b"].sort.join(",")
=> "a:b,foo:bar"
irb(main):003:0> key = ["some_name", tags, true, :c]
=> ["some_name", "a:b,foo:bar", true, :c]
irb(main):004:0> cache[key] = 1
=> 1
irb(main):005:0> cache[key] += 1
=> 2
irb(main):008:0> cache.keys.first
=> ["some_name", "a:b,foo:bar", true, :c]
```

(Technically, the object must be "hashable", but most things in ruby are. For something like an array or hash, the `.hash` method out of the box determines the hash in such a way as to be useful in this circumstance)

lib/statsd/instrument/aggregator.rb
pedro-stanaka: After reading the comment here, I went on to search and [read here](https://blog.ycrash.io/problems-with-finalizer/) that it can slow down GC activity as well.

lib/statsd/instrument/client.rb
```diff
@@ -324,11 +360,15 @@ def latency(name, sample_rate: nil, tags: nil, metric_type: nil, no_prefix: fals
...
           stop = Process.clock_gettime(Process::CLOCK_MONOTONIC, :float_millisecond)
 
-          sample_rate ||= @default_sample_rate
-          if sample_rate.nil? || sample?(sample_rate)
-            metric_type ||= datagram_builder(no_prefix: no_prefix).latency_metric_type
-            latency_in_ms = stop - start
-            emit(datagram_builder(no_prefix: no_prefix).send(metric_type, name, latency_in_ms, sample_rate, tags))
+          latency_in_ms = stop - start
+          if @enable_aggregation
+            @aggregator.aggregate_timing(name, latency_in_ms, tags: tags, no_prefix: no_prefix)
```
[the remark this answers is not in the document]
pedro-stanaka: Thx, good catch.

lib/statsd/instrument/aggregator.rb
```diff
@@ -0,0 +1,246 @@
...
+      end
+
+      attr_reader :mutex, :aggregation_state, :sink
+
+      def tags_sorted(tags)
+        return [].freeze if tags.nil? || tags.empty?
+
+        if tags.is_a?(Hash)
+          sorted = tags.sort_by { |k, _v| k.to_s }
+          tags = sorted.map! { |k, v| "#{k}:#{v}" }
```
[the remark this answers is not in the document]
pedro-stanaka: Ended up inlining the method calls and calling `map!` anyway to avoid allocs.

lib/statsd/instrument/aggregator.rb
pedro-stanaka: I put this whole method under synchronization, but now I am thinking it might be very expensive. Let's see.

Gemfile
```diff
@@ -11,3 +11,4 @@ gem "yard"
 gem "rubocop", ">= 1.0"
 gem "rubocop-shopify", require: false
 gem "benchmark-ips"
+gem "mutex_m"
```
casperisfine, line 14: Please don't use `mutex_m` it's pointless and expose public methods. Just use a simple mutex as an ivar.

lib/statsd/instrument/aggregator.rb
```diff
@@ -0,0 +1,260 @@
...
+              agg.tags,
+            )
+          else
+            StatsD.logger.error { "[#{self.class.name}] Unknown aggregation type: #{agg.type}" }
+          end
+        end
+        @aggregation_state.clear
+      end
+
+      attr_reader :mutex, :aggregation_state, :sink
```
casperisfine, line 206: I'd suggest to directly use ivars, it's better for YJIT.

lib/statsd/instrument/aggregator.rb
```diff
@@ -0,0 +1,260 @@
...
+
+        tags = tags_sorted(tags)
+        key = packet_key(name, tags, no_prefix, COUNT)
+
+        mutex.synchronize do
+          unless aggregation_state.key?(key)
+            aggregation_state[key] = AggregationValue.new(name, COUNT, 0, tags, no_prefix)
+          end
+          aggregation_state[key].value += value
+        end
```
casperisfine, line 129: What happens if someone makes a mistake and use the same `name` for two different types?

lib/statsd/instrument/aggregator.rb
```diff
@@ -0,0 +1,260 @@
...
+        @name = name
+        @type = type
+        @value = value
+        @tags = tags
+        @no_prefix = no_prefix
+      end
+    end
+
+    class Aggregator
+      CONST_SAMPLE_RATE = 1.0
```
casperisfine, line 23: I don't understand why you need to disregard sampling. You could perfectly handle it AFAICT.

lib/statsd/instrument/aggregator.rb
```diff
@@ -0,0 +1,260 @@
...
+          end
+        end
+        @aggregation_state.clear
+      end
+
+      attr_reader :mutex, :aggregation_state, :sink
+
+      # @return Array<String>
+      def tags_sorted(tags)
+        return [].freeze if tags.nil? || tags.empty?
```
casperisfine, line 210: If you're going to return a frozen aaray, you can constantize it to avoid an allocation.

lib/statsd/instrument/aggregator.rb
```diff
@@ -0,0 +1,260 @@
...
+      # @param no_prefix [Boolean] If true, the metric will not be prefixed.
+      # @return [void]
+      def increment(name, value = 1, tags: [], no_prefix: false)
+        unless thread_healthcheck
+          sink << datagram_builder(no_prefix: no_prefix).c(name, value, CONST_SAMPLE_RATE, tags)
+          return
+        end
+
+        tags = tags_sorted(tags)
+        key = packet_key(name, tags, no_prefix, COUNT)
```
casperisfine, line 122: This sorting and tag serialization is still duplicated work as we'll need to do it again when building the datagram.

Gemfile
casperisfine: Yeah, you need to update minitest.

lib/statsd/instrument/aggregator.rb
pedro-stanaka: Using POROs now, please check if that is okay.

lib/statsd/instrument/aggregator.rb
pedro-stanaka: Going to leave this for last. Would be nice do this, but I need to normalize and make sure my separator is not present in the final string parts. I also was thinking of sharing the method from `DatagramBuilder#compile_tags` or making it a class method so I can use it for this.

lib/statsd/instrument/aggregator.rb
pedro-stanaka: Yeah, I kind of agree as well, on the other hand sampling is probabilistic. There is a discussion about this here: https://github.com/Shopify/statsd-instrument/pull/366#discussion_r1635323269. More notably, datadog uses the same strategies on their Ruby and Golang clients.

lib/statsd/instrument/aggregator.rb
pedro-stanaka: Other OSS projects like statsd_exporter, simply ignore samples that come with same name and different types, we could do the same. I will write a test to document this.

lib/statsd/instrument/aggregator.rb
pedro-stanaka: Reworked this part, kept only type, no_prefix to avoid type conversions (string to symbol, string to bool) and `value`, obviously.

lib/statsd/instrument/aggregator.rb
```diff
@@ -0,0 +1,265 @@
...
+
+        @mutex.synchronize do
+          unless @aggregation_state.key?(key)
+            @aggregation_state[key] = AggregationValue.new(COUNT, 0, no_prefix)
+          end
+          @aggregation_state[key].value += value
```
casperisfine, line 132:
```suggestion
          aggregation = @aggregation_state[key] ||= AggregationValue.new(COUNT, 0, no_prefix)
          aggregation.value += value
```

lib/statsd/instrument/aggregator.rb
```diff
@@ -0,0 +1,265 @@
...
+      end
+
+      private
+
+      # @return Array<String>
+      EMPTY_ARRAY = [].freeze
+
+      def do_flush
+        @aggregation_state.each do |key, agg|
+          key_parts = key.split("|".b)
```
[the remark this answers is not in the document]
casperisfine: Also it's a bit wasteful, we could make the key a hashable PORO so you don't need to split again.

lib/statsd/instrument/aggregator.rb
```diff
@@ -0,0 +1,265 @@
...
+          tags = tags.sort_by { |k, _v| k.to_s }.map! { |k, v| "#{k}:#{v}" }
+        else
+          tags.sort!
+        end
+        DatagramBuilder.normalize_tags(tags)
+      end
+
+      def packet_key(name, tags = "".b, no_prefix = false, type = COUNT)
+        "#{DatagramBuilder.normalize_string(name)}|#{tags}|#{no_prefix}|#{type}".b
+      end
```
casperisfine, line 228: Create a `AggregationKey` PORO that keep these things as independent objects as ivars, and define `hash` and `eql?` so you can use it as hash key.

pedro-stanaka: Ah cool, I did not get this idea from the beginning. This sounds like a good approach.

lib/statsd/instrument/aggregator.rb
```diff
@@ -0,0 +1,265 @@
...
+
+      # @return Array<String>
+      EMPTY_ARRAY = [].freeze
+
+      def do_flush
+        @aggregation_state.each do |key, agg|
+          key_parts = key.split("|".b)
+          name = key_parts[0]
+          tags = key_parts[1]
+          tags = tags.empty? ? EMPTY_ARRAY : tags.split(",".b)
```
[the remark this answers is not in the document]
pedro-stanaka: Fixed it, it was supposed to be just `tags`, but I am using the value from the key now.

pedro-stanaka:
@ahayworth @casperisfine I have updated the PR addressing most, if not all, points from your reviews. Let me know what you think of this version. Going to wait for #368 so I can benchmark this properly.

lib/statsd/instrument/aggregator.rb
```diff
@@ -0,0 +1,260 @@
...
+        [@name, @tags, @no_prefix, @type].hash
+      end
+
+      def eql?(other)
+        other.is_a?(self.class) &&
+          @name == other.name &&
+          @tags == other.tags &&
+          @no_prefix == other.no_prefix &&
+          @type == other.type
+      end
```
casperisfine, line 25:
```suggestion
      def ==(other)
        other.is_a?(self.class) &&
          @name == other.name &&
          @tags == other.tags &&
          @no_prefix == other.no_prefix &&
          @type == other.type
      end
      alias_method :eql?, :==
```

lib/statsd/instrument/aggregator.rb
casperisfine: > Using objects for keys is a bit slower than using strings

Yeah, because the C implementation has to call back into Ruby. Once thing you could try is to make `AggregationKey` a `Struct`, this way it wouldn't need to call back in Ruby to hash the value.

casperisfine: Another trick you can do, is since you know `AggregationKey` is immutable, you can precompute the hash code in `initialize` and make `hash` a simple `attr_reader`.

pedro-stanaka: Yeah, did that already. Something that also made a difference was freezing the object.

test/aggregator_test.rb
```diff
@@ -0,0 +1,142 @@
+# frozen_string_literal: true
+
+require "test_helper"
+
+class Aggre < Minitest::Test
```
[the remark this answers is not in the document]
pedro-stanaka: fixed this

lib/statsd/instrument/aggregator.rb
```diff
@@ -0,0 +1,261 @@
...
+      def initialize(name, tags, no_prefix, type)
+        @name = name
+        @tags = tags
+        @no_prefix = no_prefix
+        @type = type
+      end
+
+      def hash
+        [@name, @tags, @no_prefix, @type].hash
+      end
```
ahayworth, line 17: As @casperisfine pointed out elsewhere, this is another spot for optimization (if it is needed) - we could precompute `hash` in the `initialize` method and make this just an `attr_reader`.

```ruby
attr_reader :name, :tags, :no_prefix, :hash

def initialize(name, tags, no_prefix, type)
  @name = name
  ...
  @hash = [@name, @tags, @no_prefix, @type].hash
end
```

If we see `hash` showing up in profiles, it may be worth it. But if we want to wait and see how it performs first, I think that's okay too.

lib/statsd/instrument/aggregator.rb
```diff
@@ -0,0 +1,261 @@
...
+        end
+      end
+
+      def flush
+        @mutex.synchronize { do_flush }
+      end
+
+      private
+
+      # @return Array<String>
```
ahayworth, line 178:
```suggestion
```

Seems like maybe leftovers from refactor.

lib/statsd/instrument/aggregator.rb
```diff
@@ -0,0 +1,261 @@
...
+
+      def tags_sorted(tags)
+        return "" if tags.nil? || tags.empty?
+
+        if tags.is_a?(Hash)
+          tags = tags.sort_by { |k, _v| k.to_s }.map! { |k, v| "#{k}:#{v}" }
+        else
+          tags.sort!
+        end
+        DatagramBuilder.normalize_tags(tags)
```
ahayworth, line 221: IIRC, calls to `normalize_tags` were inlined elsewhere in the codebase as an optimization - we could keep that knowledge handy in case this shows up as a hot spot in profiles.

lib/statsd/instrument/environment.rb
```diff
@@ -105,6 +105,14 @@ def statsd_batch_statistics_interval
         ))
       end
 
+      def experimental_aggregation_enabled?
+        env.key?("STATSD_ENABLE_AGGREGATION")
+      end
+
+      def aggregation_interval
+        Float(env.fetch("STATSD_AGGREGATION_INTERVAL", 2.0))
```
ahayworth, line 113: The initializer for the `Client` sets this to `5.0`; should we harmonize them?

ahayworth (approved):
Overall, this looks pretty good to me. Like I mentioned in chat, I think we should try this out on a few other big clusters before we merge - give it a really good stress-test. But, I feel good about it.

A few minor nits in comments, but nothing I think I feel like blocking over.

One question I had though - is there any value-packing supported for histograms? I noticed we don't call the aggregator for that at all. It's not used heavily internally as far as I can tell, but it did jump out at me.

lib/statsd/instrument/aggregator.rb
pedro-stanaka: I did this on a branch where i was doing benchmarking and forgot to pull it here. Commited this now.

casperisfine:
If I change the benchmark to also include the key allocation (which is more realistic)

```ruby
Benchmark.ips do |x|
  x.report("PORO") { registry[AggregationKey.new("some.key", "some_tag:1", false, :c)] }
  x.report("precomputed") { registry[PrecomputedAggregationKey.new("some.key", "some_tag:1", false, :c)] }
  x.report("Struct") { registry[StructKey.new("some.key", "some_tag:1", false, :c)] }
  x.compare!(order: :baseline)
end
```

```
$ ruby /tmp/hash.rb
ruby 3.3.3 (2024-06-12 revision f1c7b6f435) [arm64-darwin23]
                PORO   276.197k i/100ms
         precomputed   327.271k i/100ms
              Struct   415.442k i/100ms
                PORO      2.816M (± 1.4%) i/s -     14.086M in   5.003790s
         precomputed      3.350M (± 0.9%) i/s -     17.018M in   5.080119s
              Struct      4.154M (± 1.5%) i/s -     21.188M in   5.101720s

Comparison:
                PORO:  2815622.0 i/s
              Struct:  4153942.4 i/s - 1.48x  faster
         precomputed:  3350230.3 i/s - 1.19x  faster

$ ruby --yjit /tmp/hash.rb
ruby 3.3.3 (2024-06-12 revision f1c7b6f435) +YJIT [arm64-darwin23]
                PORO   407.112k i/100ms
         precomputed   429.832k i/100ms
              Struct   434.237k i/100ms
                PORO      4.383M (± 1.3%) i/s -     21.984M in   5.016207s
         precomputed      4.692M (± 1.0%) i/s -     23.641M in   5.039512s
              Struct      4.489M (± 4.8%) i/s -     22.580M in   5.045144s

Comparison:
                PORO:  4383350.0 i/s
         precomputed:  4691568.5 i/s - 1.07x  faster
              Struct:  4489005.9 i/s - same-ish: difference falls within error
```

This would suggest to use `Struct`.

casperisfine:
> SERIES_COUNT=5
> UDP Batched: 427,910.0 events/s, Total time: 2.1s
> UDP Batched with Aggregation: 510,489.4 events/s, Total time: 1.76s

TBH I was expecting a much bigger difference is such favorable scenario. Also surprised to see the number don't change that much when cardinality changes for either implementation. I'd be curious to see a profile of that benchmark.

pedro-stanaka:
> TBH I was expecting a much bigger difference is such favorable scenario. Also surprised to see the number don't change that much when cardinality changes for either implementation. I'd be curious to see a profile of that benchmark.

The emitting threads are just dominated by the events which are not aggregated:

image


Here is the Vernier file:  

[benchmark_profile_udp_async_aggregated.json](https://github.com/user-attachments/files/16176656/benchmark_profile_udp_async_aggregated.json)

pedro-stanaka:
I am not going to use Struct for now, it is actually making performance worse:

Pre computed object:
```
▶ RUBY_YJIT_ENABLE=true ITERATIONS=6000 SERIES_COUNT=5 THREAD_COUNT=5 ./benchmark/local-udp-throughput                                                                                                                                                                                                               
===== UDP batched throughput (5 threads) - total events: 900000 =====
events: 433,375.1/s. Total time: 2.08s
===== UDP batched with aggregation throughput (5 threads) - total events: 900000 =====
events: 510,911.1/s. Total time: 1.76s
===== UDP batched with aggregation and 5 second interval throughput (5 threads) - total events: 900000 =====
events: 511,985.3/s. Total time: 1.76s

```

Struct:
```
▶ RUBY_YJIT_ENABLE=true ITERATIONS=6000 SERIES_COUNT=5 THREAD_COUNT=5 ./benchmark/local-udp-throughput                                                                                                                                                                                                                   
===== UDP sync throughput (5 threads) - total events: 900000 =====
events: 134,456.6/s. Total time: 6.69s
===== UDP batched throughput (5 threads) - total events: 900000 =====
events: 427,699.9/s. Total time: 2.1s
===== UDP batched with aggregation throughput (5 threads) - total events: 900000 =====
events: 475,442.3/s. Total time: 1.89s
===== UDP batched with aggregation and 5 second interval throughput (5 threads) - total events: 900000 =====
events: 480,398.2/s. Total time: 1.87s
```

ahayworth:
> @ahayworth we're indeed value-packing distributions/histograms. This is done in these places:

@pedro-stanaka I meant that I see no calls to the aggregator from the `Client#histogram` method - I think maybe it was just missed?

pedro-stanaka:
> > @ahayworth we're indeed value-packing distributions/histograms. This is done in these places:
> 
> @pedro-stanaka I meant that I see no calls to the aggregator from the `Client#histogram` method - I think maybe it was just missed?

Duh, my bad. Yeah, I missed that one. Thanks for pointing it out.

pedro-stanaka:
> Like, either something isn't working as intended, or either the benchmark isn't doing what I think, or either this is a ton of extra complexity for not a whole lot of performance gain.

DogStatsD is big mess TBH, this PR concerns aggregating metrics, but in our benchmark we are also sending all the other type of "events" that just Datadog Agent can understand like service checks and events. 

If I change the benchmark to just send metrics, the results are dramatically different:

```
✗ RUBY_YJIT_ENABLE=true ITERATIONS=6000 SERIES_COUNT=5 THREAD_COUNT=5 ./benchmark/local-udp-throughput                                                                                                                                                                                                🕙 15:00:59 
===== UDP sync throughput (5 threads) - total events: 900000 =====
events: 133,204.5/s. Total time: 6.76s
===== UDP batched throughput (5 threads) - total events: 900000 =====
events: 450,433.1/s. Total time: 2.0s
===== UDP batched with aggregation throughput (5 threads) - total events: 900000 =====
events: 836,921.2/s. Total time: 1.08s
===== UDP batched with aggregation and 5 second interval throughput (5 threads) - total events: 900000 =====
events: 837,549.0/s. Total time: 1.07s
```

casperisfine:
> but in our benchmark we are also sending all the other type of "events" that just Datadog Agent can understand like service checks and events.
> 
> If I change the benchmark to just send metrics, the results are dramatically different:

Ah, just remove that from the benchmark.

jmreid:
FYI: We've deployed `8a98b8fdc0d0308dccf0173afa73f063fbb4a314` to SFR and it has had a substantial impact:
image

We're roll it out everywhere because we're facing some node CPU constraints and this reduction is too valuable to pass up. We will switch to the released version once this PR merges.

pedro-stanaka:
@casperisfine do you have any other remarks here? Can I proceed with merge?

lib/statsd/instrument/aggregator.rb
```diff
@@ -0,0 +1,257 @@
...
+        end
+
+        tags = tags_sorted(tags)
+        key = packet_key(name, tags, no_prefix, type)
+
+        @mutex.synchronize do
+          if @aggregation_state.key?(key) && @aggregation_state[key].size + 1 >= @max_values
+            do_flush
+          end
+          @aggregation_state[key] ||= []
```
casperisfine, line 150:
```suggestion
          values = @aggregation_state[key] ||= []
          if values.size + 1 >= @max_values
            do_flush
          end
```

lib/statsd/instrument/datagram_builder.rb
```diff
@@ -25,31 +59,35 @@ def initialize(prefix: nil, default_tags: nil)
...
       end
 
       def s(name, value, sample_rate, tags)
-        generate_generic_datagram(name, value, "s", sample_rate, tags)
+        generate_generic_datagram(name, [value], "s", sample_rate, tags)
       end
 
       def h(name, value, sample_rate, tags)
-        generate_generic_datagram(name, value, "h", sample_rate, tags)
+        generate_generic_datagram(name, [value], "h", sample_rate, tags)
```
casperisfine, line 78: Passing an array here is an extra allocation for every call, wouldn't it be simpler to do the `.join(":")` in `timing_value_packed ` and leave `generate_generic_datagram` unchanged?

lib/statsd/instrument/datagram_builder.rb
```diff
@@ -88,27 +126,7 @@ def generate_generic_datagram(name, value, type, sample_rate, tags)
...
-            buffer << key << ":" << value
-          end
-        else
-          if tags.any? { |tag| /[|,]/.match?(tag) }
-            tags = tags.map { |tag| tag.tr("|,", "") }
-          end
-          buffer << tags.join(",")
-        end
-        buffer
+        DatagramBuilder.normalize_tags(tags, buffer)
```
casperisfine, line 129: If the method is just a delegator, might as well get rid of it and call the original one directly.

casperisfine: I'm afraid I don't understand the question.

pedro-stanaka: The DatagramBuilder had a `protected` method `compile_tags` which comes from the "abstract" class StatsDDatagramBuilder. I wanted a `public` method that does exactly what `compile_tags` does. 

Should I just make the method public in `StatsDDatagramBuilder`?

pedro-stanaka: I am going with the solution of creating a new alias for now.

casperisfine: I don't get why you need a public one but ok.

pedro-stanaka: I am using this from the Aggregator to create the key, so I need to share it. We can iterate later. Thx for all the help and feedback

PR #372, opened 2024-08-02
UDS Sink + async UDS sink batched

pedro-stanaka:
## Summary

Using UDS (Unix Domain Sockets) is one of the main recommendation from Datadog to achieve higher throughput for StatsD traffic. So far, this client did not provide such a sink. 

In this PR I am introducing a UDS sink that uses same batching logic from UDP sink.


### Follow up improvements

* https://github.com/Shopify/statsd-instrument/pull/374 - Here I am removing the specific types of sinks and making them generic, then I introduce the concept of `Connection` which will then interact with the sockets. This solves the duplication problem I introduced in this PR.

Ended up merging this PR above and squashing the other commits, when reviewing, one can review the commits separately. 


### Benchmarks
We can see from benchmarks that using larger batches improves client throughput with a considerable margin:

```
▶ RUBY_YJIT_ENABLE=true ./benchmark/local-udp-throughput                                                                                                                                                                                                                                             
Logs are stored in /var/folders/9v/vqq8s1cd2n53w68813dwjz4c0000gn/T/statsd-instrument-benchmarks
===== UDP sync throughput (5 threads) - total events: 150000 =====
events: 138,061.4/s
===== UDP batched throughput (5 threads) - total events: 150000 =====
events: 734,325.8/s
===== UDS batched with small packet throughput (5 threads) - total events: 150000 =====
events: 1,386,411.3/s
===== UDS batched with jumbo packet throughput (5 threads) - total events: 150000 =====
events: 1,248,824.0/s

```

### DIFF
--- a/lib/statsd/instrument/batched_sink.rb
+++ b/lib/statsd/instrument/batched_sink.rb
@@ -1,22 +1,30 @@
 # frozen_string_literal: true
 
+require "forwardable"
+
 module StatsD
   module Instrument
-    # @note This class is part of the new Client implementation that is intended
-    #   to become the new default in the next major release of this library.
-    class BatchedUDPSink
+    class BatchedSink
+      extend Forwardable
+
+      def_delegator :@sink, :host
+      def_delegator :@sink, :port
+
       DEFAULT_THREAD_PRIORITY = 100
       DEFAULT_BUFFER_CAPACITY = 5_000
       # https://docs.datadoghq.com/developers/dogstatsd/high_throughput/?code-lang=ruby#ensure-proper-packet-sizes
       DEFAULT_MAX_PACKET_SIZE = 1472
       DEFAULT_STATISTICS_INTERVAL = 0 # in seconds, and 0 implies disabled-by-default.
 
-      attr_reader :host, :port
-
       class << self
         def for_addr(addr, **kwargs)
-          host, port_as_string = addr.split(":", 2)
-          new(host, Integer(port_as_string), **kwargs)
+          if addr.include?(":")
+            sink = StatsD::Instrument::Sink.for_addr(addr)
+            new(sink, **kwargs)
+          else
+            connection = UdsConnection.new(addr)
+            new(connection, **kwargs)
+          end
         end
 
         def finalize(dispatcher)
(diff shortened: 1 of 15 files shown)


lib/statsd/instrument/batched_sink.rb
```diff
@@ -153,15 +168,23 @@ def initialize(host, port, buffer_capacity, thread_priority, max_packet_size, st
...
           if statistics_interval > 0
-            @statistics = DispatcherStats.new(statistics_interval)
+            type = case @sink.connection
+            when UdsConnection
+              "uds"
+            when UdpConnection
+              "udp"
+            else
+              "unknown"
+            end
```
casperisfine, line 178: Might as well add a `type` method on `Connection`, the `Sink` shouldn't have to know about the possible implementations.

lib/statsd/instrument/sink.rb
```diff
@@ -0,0 +1,97 @@
...
+
+      def connection
+        thread_store[object_id] ||= @connection
+      end
+
+      def host
+        if connection.respond_to?(:host)
+          connection.host
+        else
+          ""
```
casperisfine, line 74: Should be `nil`

lib/statsd/instrument/sink.rb
```diff
@@ -0,0 +1,97 @@
...
+        end
+      end
+
+      def port
+        if connection.respond_to?(:port)
+          connection.port
+        else
+          0
```
casperisfine, line 82: Should be `nil`.

lib/statsd/instrument/sink.rb
```diff
@@ -0,0 +1,97 @@
...
+      def connection_type
+        connection.class.name
+      end
+
+      def connection
+        thread_store[object_id] ||= @connection
+      end
+
+      def host
+        if connection.respond_to?(:host)
```
casperisfine, line 71: Also not great to check `respond_to?`, it's better to define `#host` and `#port` on `UDSConnection` to return `nil`.

lib/statsd/instrument/uds_connection.rb
```diff
@@ -0,0 +1,40 @@
...
+            "[StatsD::Instrument::UdsConnection] max_packet_size must be greater than 0, " \
+              "using default: #{DEFAULT_MAX_PACKET_SIZE}"
+          end
+        end
+        @socket_path = socket_path
+        @max_packet_size = max_packet_size
+      end
+
+      def send_datagram(message)
+        socket.sendmsg_nonblock(message)
```
casperisfine, line 20: The `_nonblock` version of IO methods raise an exception if the socket can't be written into and you're suppose to retrt later: https://docs.ruby-lang.org/en/3.3/BasicSocket.html#method-i-sendmsg_nonblock

So this doesn't seem correct.

If you want to just drop the packet if the socket is full, then you can dismiss the exception with `exception: false` and `:wait_writable` will be returned in case the socket is full.

lib/statsd/instrument/batched_sink.rb
pedro-stanaka: Did that. Thanks!

PR #373, opened 2024-08-07
Deprecate Ruby 2.6, improve CI setup

pedro-stanaka:
## Summary

Ruby 2.6 reached its EOL in March 2022, in newer versions of the library we could disable support for that deprecated ruby version.
Also, our CI for github was using actions that were running with deprecated versions of NodeJS, I am upgrading the actions versions so we can comply with the newer Node versions.
Lastly, I am updating the versions we run tests against for JRuby and TruffleRuby

- **abandon support for 2.6**
- **Using newer version of checkout action get rid of deprecation warnings;**
- **update jruby and truffleruby versions**

### DIFF
--- a/.github/workflows/benchmark.yml
+++ b/.github/workflows/benchmark.yml
@@ -8,7 +8,7 @@ jobs:
     runs-on: ubuntu-latest
 
     steps:
-    - uses: actions/checkout@v3
+    - uses: actions/checkout@v4
 
     - name: Set up Ruby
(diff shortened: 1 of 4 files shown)


casperisfine:
> * abandon support for 2.6

If it allows to simplify some code I'm all for it, but if it doesn't I see no reason to bump the requirement.

casperisfine (approved):
Please squash your commits. Other than that 👍

PR #384, opened 2024-10-28
Aggregation: fixing bug when sending metrics synchronously

pedro-stanaka:
## ✅ What

I am solving a problem when we try to send metrics and the main Thread is either dying or dead. In those cases, the library will try to send everything synchronously. But we were not covering that path with tests and had a call to `sink` without `@` and it was throwing errors.

## 🤔 Why

This is currently causing problems in production, in certain contexts when the pod is shutting down for example.

## 👩🔬 How to validate

I added tests that actually cover that portion of the code, even thought it does not use some trapping or actual killing of the main thread, it should at least exercise the code.

## Checklist

### DIFF
--- a/lib/statsd/instrument/aggregator.rb
+++ b/lib/statsd/instrument/aggregator.rb
@@ -252,6 +252,8 @@ module StatsD
 
             # If the PID changed, the process forked, reset the aggregator state
             if @pid != Process.pid
+              # TODO: Investigate/replace this with Process._fork hook.
+              # https://github.com/ruby/ruby/pull/5017
               StatsD.logger.debug do
                 "[#{self.class.name}] Restarting the flush thread after fork. State size: #{@aggregation_state.size}"
               end


andrejbranch (approved):

lib/statsd/instrument/aggregator.rb
```diff
@@ -243,15 +243,21 @@ def datagram_builder(no_prefix:)
...
         @mutex.synchronize do
           unless @flush_thread&.alive?
+            # The main thread is dead, fallback to direct writes
             return false unless Thread.main.alive?
 
+            # The main thread forked, reset the aggregator state
             if @pid != Process.pid
               StatsD.logger.debug { "[#{self.class.name}] Restarting the flush thread after fork" }
+              # Try one last flush to ensure all metrics are sent
+              do_flush
```
ahayworth, line 253: Hm. I wonder if there's any chance that the flush thread of the parent process gets called after it `fork`s off, thereby resulting in dual writes of the same data point?

...and _if_ that's something that can happen, are there any downsides that we might want to consider? I worry it might cause counter increments to get double-counted, that kind of thing.

However, I haven't really thought too deeply about it, so if that's all impossible or nothing to worry about, then 🎉 this is nice.

test/aggregator_test.rb
```diff
@@ -139,4 +159,89 @@ def test_with_prefix
...
+  def test_synchronous_operation_on_thread_failure
+    # Force thread_healthcheck to return false
+    @subject.stubs(:thread_healthcheck).returns(false)
+
+    @subject.increment("foo", 1, tags: { foo: "bar" })
+    @subject.aggregate_timing("bar", 100, tags: { foo: "bar" })
+    @subject.gauge("baz", 100, tags: { foo: "bar" })
+
+    # Verify metrics were sent immediately
+    assert_equal(3, @sink.datagrams.size)
```
ahayworth, line 172: This might be paranoid, but perhaps it's also worth asserting that no methods were called on the `Aggregator`'s internal `@aggregation_state`. Otherwise the test might be a little racy, I'm not sure.

I may be overthinking that. 🤷

ahayworth (commented):
Overall the fix is straightforward, but the added `do_flush` gives me a bit of pause. I won't block over it though, because I haven't really though too deeply about that.

I also have a minor worry about the robustness of the test, but again - I might be way off base with that concern, so it's not worth blocking over.

❤️

test/aggregator_test.rb
pedro-stanaka: Added assertions to be on the safe side.

pedro-stanaka:
Tested in staging for a production app. Looks good.

PR #391, opened 2024-12-18
Fix UDS socket configuration and improve error handling

pedro-stanaka:
### Problem

When using Unix Domain Sockets (UDS) with statsd-instrument, the `max_packet_size` configuration wasn't being properly passed through from the environment to the UDS connection. Additionally, error handling and debugging information for socket errors needed improvement.

### What was done

1. Fixed the `Environment#default_sink_for_environment` method to properly pass `max_packet_size` to both UDS and UDP connections
2. Added better error handling and debugging information in the Sink class
3. Added comprehensive tests for UDS socket configuration
4. Fixed socket mock expectations in UDP sink tests

## Notes
When using UDS with aggregation, be aware that the OS might impose its own limits on datagram sizes. The socket buffer size can be configured, but there might still be per-message size limits that can't be exceeded.

## 🤔 Why

The send buffer should match the max_packet size on the BatchedSink, otherwise messages will be dropped.

## 👩🔬 How to validate

Added tests for the behavior.

## Checklist

#gsd:41649

### DIFF
--- a/test/environment_test.rb
+++ b/test/environment_test.rb
@@ -78,6 +78,7 @@ class EnvironmentTest < Minitest::Test
   end
 
   def test_client_from_env_uses_uds_sink_with_correct_packet_size_in_production
+    skip_on_jruby("JRuby does not support UNIX domain sockets")
     socket_path = "/tmp/statsd-test-#{Process.pid}.sock"
 
     # Create a UDS server socket
@@ -105,6 +106,7 @@ class EnvironmentTest < Minitest::Test
   end
 
   def test_client_from_env_uses_default_packet_size_for_uds_when_not_specified
+    skip_on_jruby("JRuby does not support UNIX domain sockets")
     socket_path = "/tmp/statsd-test-#{Process.pid}-default.sock"
 
     # Create a UDS server socket
@@ -134,6 +136,7 @@ class EnvironmentTest < Minitest::Test
   end
 
   def test_client_from_env_uses_batched_uds_sink_with_correct_packet_size
+    skip_on_jruby("JRuby does not support UNIX domain sockets")
     socket_path = "/tmp/statsd-test-#{Process.pid}-batched.sock"
 
     # Create a UDS server socket
--- a/test/test_helper.rb
+++ b/test/test_helper.rb
@@ -30,6 +30,15 @@ module StatsD
   end
 end
 
+# Add helper methods available to all tests
+module Minitest
+  class Test
+    def skip_on_jruby(message = "Test skipped on JRuby")
+      skip(message) if RUBY_ENGINE == "jruby"
+    end
+  end
+end
+
 StatsD.logger = Logger.new(File::NULL)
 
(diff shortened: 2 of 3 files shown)


lib/statsd/instrument/environment.rb
```diff
@@ -140,19 +140,25 @@ def default_sink_for_environment
...
+              statsd_socket_path,
+              max_packet_size: statsd_max_packet_size.to_i,
+            )
           else
             host, port = statsd_addr.split(":")
-            StatsD::Instrument::UdpConnection.new(host, port.to_i)
+            StatsD::Instrument::UdpConnection.new(
+              host,
+              port.to_i,
+              max_packet_size: statsd_max_packet_size.to_i,
```
ahayworth, line 152: I think this `to_i` is redundant, no? It looks to me like `statsd_max_packet_size` always returns an integer according to how you've defined it above.

lib/statsd/instrument/udp_connection.rb
```diff
@@ -30,6 +31,7 @@ def type
       def socket
         @socket ||= begin
           socket = UDPSocket.new
+          socket.setsockopt(Socket::SOL_SOCKET, Socket::SO_SNDBUF, @max_packet_size)
```
ahayworth, line 34: Like you mentioned in the PR description, there's a chance that the user could set a value higher than what the operating system would support. If that happens, wouldn't we still be liable to unexpectedly lose data in the same way the PR describes?

I think it _might_ be worth validating the `@max_packet_size` somehow (probably before we get down to this point, but 🤷 wherever it makes the most sense). I'm not sure if it's best to manually (or programmatically, somehow) check what the actual OS limits are and validate configuration accordingly, or possibly just check and see if we're able to set whatever the user asked for (and then take action accordingly: log a big-ol' warning, lower the max packet size to whatever the buffer is actually set to at this point, run around with our hair on fire, etc).

It's pretty easy to ask the socket what that value is before/after you try and modify it:

```ruby
irb(main):001> require 'socket'
=> true
irb(main):002> s = UDPSocket.new
=> #<UDPSocket:fd 6, AF_INET, 0.0.0.0, 0>
irb(main):003> s.getsockopt(Socket::SOL_SOCKET, Socket::SO_SNDBUF).int
=> 9216
irb(main):004> s.setsockopt(Socket::SOL_SOCKET, Socket::SO_SNDBUF, 99_999_999)
=> 0
irb(main):005> s.getsockopt(Socket::SOL_SOCKET, Socket::SO_SNDBUF).int
=> 8388608
irb(main):006>
```

Of course, IDK if this works the same way on Windows, or if that matters much to us.

ahayworth (approved):
I left some thought about validation of the value, but overall this looks okay to me.

lib/statsd/instrument/udp_connection.rb
pedro-stanaka: That is an awesome call out Andrew! ❤️ Will change it to do so.

PR #393, opened 2025-02-03
introduce an opinionatedly better statsd helper that gives you access…

raginpirate:
… to all your variables

## Checklist

### DIFF
--- a/lib/statsd/instrument.rb
+++ b/lib/statsd/instrument.rb
@@ -38,10 +38,18 @@ module StatsD
       # Generates the tags for an instrumented method.
       # @private
       # @return [Array[String]]
-      def generate_tags(tags, callee, *args)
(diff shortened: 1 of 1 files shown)


lib/statsd/instrument.rb
```diff
@@ -38,10 +38,18 @@ def generate_metric_name(name, callee, *args)
       # Generates the tags for an instrumented method.
       # @private
       # @return [Array[String]]
-      def generate_tags(tags, callee, *args)
+      def generate_tags(tags, callee, result=nil, *args)
```
raginpirate, line 41: I need to fix this, probably requires a bool `use_result = false` for real nilable result objects.

raginpirate:
Closing this as I'm opinionated that core needs a better metrics solution now.

PR #396, opened 2025-03-25
Memoize tags to avoid recomputing

dalehamel:
## ✅ What

Try to avoid optimize tag computation by memoizing tag string, so that identical inputs will use the cached value rather than recompute.

## 🤔 Why

`compile_tags` is showing up as the top frame in some profiles that emit lots of metrics, dominating execution time more than it should

Screenshot 2025-03-25 at 3 08 39 PM

## 👩🔬 How to validate

Ideally we should benchmark this to assert it actually makes things faster. I'm not sure how ruby generates cache keys for complex objects like arrays and dicts, but it seems to be based on the actual values:

```
irb(main):001> key = {"a" => :b, "b" => :c}
=> {"a" => :b, "b" => :c}
irb(main):002> hash = {key => :val}
=> {{"a" => :b, "b" => :c} => :val}
irb(main):003> hash[{"b" => :c, "a" => :b}]
=> :val
```

I'm assuming that ruby internally is doing this efficiently and that this computation is cheaper than re-computing the tags.

## Checklist

### DIFF
--- a/lib/statsd/instrument/datagram_builder.rb
+++ b/lib/statsd/instrument/datagram_builder.rb
@@ -104,30 +104,43 @@ module StatsD
       end
 
       def compile_tags(tags, buffer = "".b)
+        @tag_cache ||= {}
+        if (tag_str = @tag_cache[tags])
+          buffer << tag_str
+          return buffer
+        end
         if tags.is_a?(String)
-          tags = self.class.normalize_string(tags) if /[|,]/.match?(tags)
-          buffer << tags
+          normalized_tags = /[|,]/.match?(tags) ? self.class.normalize_string(tags) : tags
+          @tag_cache[tags] = normalized_tags
+          buffer << normalized_tags
           return buffer
         end
         if tags.is_a?(Hash)
           first = true
+          tag_repr = "".b
           tags.each do |key, value|
             if first
               first = false
             else
-              buffer << ","
+              tag_repr << ","
             end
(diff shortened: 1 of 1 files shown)


dalehamel:
linter failure is on files not touched by this PR 🤨

lib/statsd/instrument/datagram_builder.rb
```diff
@@ -104,30 +104,43 @@ def generate_generic_datagram(name, value, type, sample_rate, tags)
       end
 
       def compile_tags(tags, buffer = "".b)
+        @tag_cache ||= {}
```
dalehamel, line 107: note that the cache is only scoped to the individual datagram_builder instance, so if a new instance is used we won't benefit from memoizing.

casperisfine:
What's the eviction policy like? How do you ensure that `sometag: rand` won't grow the cache to infinity?

dalehamel:
@casperisfine @pedro-stanaka sorry for wasting your review time but ultimately the speedups here I don't think are worth the overhead of managing the cache. In the case I got called in to look at, i think that lots of metrics are being emitted and the cost of computing the cache key is only a bit better than computing the tag itself.

The issues of synchronization and cache TTL are valid and will probably further tilt the balance away from this being a worthwhile optimization.

I'm going to close this and if someone thinks of a better caching strategy, feel free to build on it.

PR #417, opened 2026-03-20
Expose the metric name for CompiledMetric

michaelherold:
## ✅ What

This change makes the metric name accessible via a top-level `.metric_name` method on the compiled metric.

## 🤔 Why

This has the benefit of making it easy to use the metric name in tests in lieu of defining constants to share the name.

## 👩🔬 How to validate

The test suite makes the usage quite clear; it's a single method on the subclass of `CompiledMetric`.

## Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,8 @@ section below.
 
 ## Unreleased changes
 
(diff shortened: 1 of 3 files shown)


pedro-stanaka (approved):

pedro-stanaka (approved):
Requesting a small change, but going to leave approved to unblock you.

PR #422, opened 2026-06-22
Fix compiled distribution sampling

### LINKED ISSUE #421: `CompiledMetric::Distribution` applies the sample rate twice
## Summary

StatsD "sampling" is a load-saving trick: instead of sending every event, the client sends only a fraction (say 1 in 100) and stamps the packet with the sample rate (`@0.01`). The backend then multiplies the counts back up (×100) to estimate the true number. The deal only works if the client rolls the dice *once*.

`CompiledMetric::Distribution` rolls the dice *twice*. A sampled distribution is therefore emitted with probability `rate²`, but the packet still says `@rate`, so the backend only scales by `1/rate`. The net effect is that a sampled distribution reads *`rate` times its true value* — e.g. a metric defined with `sample_rate: 0.001` shows up ~1000× too low.

`CompiledMetric::Counter` and `CompiledMetric::Gauge` are *not* affected (they sample once, correctly). Only `Distribution` is wrong, and only when `sample_rate < 1`. Unsampled distributions are fine.

## Root Cause

1) `Distribution` opts into latency measuring, which pulls in the extra gate (`lib/statsd/instrument/compiled_metric.rb`):
```
class Distribution < CompiledMetric
  class << self
    # ...
    def allow_measuring_latency
      true   # Counter and Gauge do NOT override this (defaults to false)
    end
  end
end
```

2) Because `allow_measuring_latency` is true, the generated emit method embeds the block handler, which samples (`generate_block_handler`). This preamble runs at the top of the method on *every* call, including the plain `.distribution(value)` (no-block) form:
```
__sample_rate__ ||= @sample_rate
if __sample_rate__ && !sample?(__sample_rate__)   # <-- sampling gate # 1
  if block_given?
    return yield
  end
  return StatsD::Instrument::VOID
end
```

3) The same value is then handed to `emit_precompiled_distribution_metric`, which samples again (`lib/statsd/instrument/client.rb`):
```
def emit_precompiled_distribution_metric(precompiled_datagram, value)
  sample_rate = precompiled_datagram.sample_rate

  if @enable_aggregation
    # Sampling decision is done at the definition of a compiled metric, see ...CompiledMetric.define
    if sample_rate.nil? || sample?(sample_rate)            # <-- sampling gate #2
      @aggregator.aggregate_precompiled_timing_metric(precompiled_datagram, value)
    end
    return StatsD::Instrument::VOID
  end

  if sample_rate.nil? || sample?(sample_rate)              # <-- sampling gate #2
    emit(precompiled_datagram.to_datagram(value))
  end

  StatsD::Instrument::VOID
end
```

For a `Counter`, `allow_measuring_latency` is false, so gate # 1 doesn't exist and `emit_precompiled_increment_metric`'s gate is the only one — correct. For a `Distribution`, both gates fire: effective emit probability is `rate²`, while the datagram carries `@rate`. The comment in `emit_precompiled_*` ("Sampling decision is done at the definition") suggests the intent was a single decision, but the gate is still applied here, duplicating gate # 1.

### Reproduction

```
class C < StatsD::Instrument::CompiledMetric::Counter
  define(name: "repro.counter", sample_rate: 0.5)
end
class D < StatsD::Instrument::CompiledMetric::Distribution
  define(name: "repro.distribution", sample_rate: 0.5)
end

100_000.times { C.increment }
100_000.times { D.distribution(1) }
```
Expected: both send ~50,000 datagrams (each stamped `@0.5`, reconstructing ~100,000). Actual: the counter sends ~50,000, but the distribution sends ~25,000 (`0.5²`), reconstructing only ~50,000 — half the true value.

### Option 1 (recommended) — keep the block-skip optimization. 

The gate in `generate_block_handler` is intentional: it lets a sampled-out call skip *executing and timing* the block. Keep it, but tell the precompiled emit the decision was already made so it doesn't re-sample.

`client.rb` — add a `pre_sampled:` flag:
```
def emit_precompiled_distribution_metric(precompiled_datagram, value, pre_sampled: false)
  sample_rate = precompiled_datagram.sample_rate

  if @enable_aggregation
    if pre_sampled || sample_rate.nil? || sample?(sample_rate)
      @aggregator.aggregate_precompiled_timing_metric(precompiled_datagram, value)
    end
    return StatsD::Instrument::VOID
  end

  if pre_sampled || sample_rate.nil? || sample?(sample_rate)
    emit(precompiled_datagram.to_datagram(value))
  end

  StatsD::Instrument::VOID
end
```
`compiled_metric.rb` — when the method already gated (i.e. `allow_measuring_latency`), pass the flag from both `define_dynamic_method` and `define_static_method`:
```
@singleton_client.emit_precompiled_#{method}_metric(__datagram__, __value__#{", pre_sampled: true" if allow_block})
```
Net result: gate # 1 makes the only sampling decision for distributions; counters/gauges keep using gate # 2. Each metric is sampled exactly once.

### Option 2 (explored, simplest) — single source of truth.

Remove the sampling gate from `generate_block_handler` entirely and let `emit_precompiled_*_metric` be the only place that samples (how counters already work). Obviously correct, at the cost of always executing and timing the block even on a sampled-out call (you lose the micro-optimization Option 1 preserves).

maxveldink:
## ✅ What

Fixes `CompiledMetric::Distribution` so sampled metrics only make one sampling decision.

`CompiledMetric::Distribution` already samples in the generated latency/block handler. The precompiled distribution client emit path then sampled again, which made sampled compiled distributions emit with probability `sample_rate²` while still sending datagrams marked with only `|@sample_rate`.

This PR routes pre-sampled compiled distributions to a dedicated private emit helper so the generated method's sampling decision is reused for both direct emission and aggregation.

Also bumps the patch version to `3.11.1` for release.

Fixes #421.

## 🤔 Why

A compiled distribution with `sample_rate: 0.001` should emit about 1 in 1000 calls and let the backend scale by 1000. Instead it emitted about 1 in 1,000,000 calls and still told the backend to scale by only 1000, under-reporting by roughly the sample rate.

Counters and gauges were not affected because they do not use the generated latency/block handler.

## 👩🔬 How to validate

Tests added coverage that fails on `main` because `sample?` is called twice:

- sampled compiled distribution value path
- sampled compiled distribution block path
- sampled compiled distribution with aggregation enabled

Validation run locally:

```text
Full test suite:
390 runs, 1098 assertions, 0 failures, 0 errors

RuboCop changed files:
4 files inspected, no offenses detected
```

## Benchmark

Temporary benchmark script based on `benchmark/send-metrics-to-dev-null-log`, focused on compiled distributions. Higher `i/s` is better.

Environment: Ruby 3.4.9, arm64-darwin24.

`main` → this branch:

```text
compiled distribution value static rate=1
2.351M -> 2.579M i/s (+9.7%)

compiled distribution value dynamic tags rate=1
1.437M -> 1.509M i/s (+5.0%)

compiled distribution value static rate=0.5 sampled-in
2.055M -> 2.364M i/s (+15.0%)

compiled distribution block static rate=1
1.669M -> 1.746M i/s (+4.6%)
```

## Checklist

### DIFF
--- a/lib/statsd/instrument/compiled_metric.rb
+++ b/lib/statsd/instrument/compiled_metric.rb
@@ -177,6 +177,10 @@ module StatsD
           default_val = default_value
           default_val_assignment = default_val.nil? ? "" : " = #{default_val.inspect}"
           allow_block = allow_measuring_latency
+          # Block-enabled compiled metrics run through generate_block_handler first,
+          # which applies sample_rate before emitting. Use the presampled client helper
+          # here so distributions don't make a second sampling decision.
+          client_emit_method = allow_block ? "emit_presampled_precompiled_#{method}_metric" : "emit_precompiled_#{method}_metric"
 
           method_code = <<~RUBY
             def self.#{method}(__value__#{default_val_assignment}, #{tag_names.map { |name| "#{name}:" }.join(", ")})
@@ -218,7 +222,7 @@ module StatsD
 
               __datagram__ ||= PrecompiledDatagram.new([#{tag_names.join(", ")}], @datagram_blueprint, @sample_rate)
 
-              @singleton_client.emit_precompiled_#{method}_metric(__datagram__, __value__)
+              @singleton_client.#{client_emit_method}(__datagram__, __value__)
               __return_value__
             end
           RUBY
@@ -233,12 +237,16 @@ module StatsD
           method = method_name
           default_val = default_value
           allow_block = allow_measuring_latency
+          # Block-enabled compiled metrics run through generate_block_handler first,
+          # which applies sample_rate before emitting. Use the presampled client helper
+          # here so distributions don't make a second sampling decision.
+          client_emit_method = allow_block ? "emit_presampled_precompiled_#{method}_metric" : "emit_precompiled_#{method}_metric"
 
           instance_eval(<<~RUBY, __FILE__, __LINE__ + 1)
             def self.#{method}(__value__ = #{default_val.inspect})
               __return_value__ = StatsD::Instrument::VOID
               #{generate_block_handler if allow_block}
-              @singleton_client.emit_precompiled_#{method}_metric(@static_datagram, __value__)
+              @singleton_client.#{client_emit_method}(@static_datagram, __value__)
               __return_value__
             end
           RUBY
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,10 @@ section below.
 
 ## Unreleased changes
 
+## Version 3.11.1
+
+- Fix `CompiledMetric::Distribution` applying `sample_rate` twice.
+
 ## Version 3.11.0
 
 - [#418](https://github.com/Shopify/statsd-instrument/pull/418) - Prevent misuse of `CompiledMetric` definitions by requiring a subclass when defining one.
--- a/lib/statsd/instrument/client.rb
+++ b/lib/statsd/instrument/client.rb
@@ -285,6 +285,25 @@ module StatsD
         StatsD::Instrument::VOID
       end
 
+      # Emits a pre-sampled compiled distribution metric.
+      # This method is used when the generated compiled metric method already made
+      # the sampling decision before measuring block latency.
+      #
+      # @param precompiled_datagram [StatsD::Instrument::CompiledMetric::PrecompiledDatagram]
+      #   The precompiled metric datagram
+      # @param value [Numeric] The metric value
+      # @return [void]
+      # @api private
+      def emit_presampled_precompiled_distribution_metric(precompiled_datagram, value)
+        if @enable_aggregation
+          @aggregator.aggregate_precompiled_timing_metric(precompiled_datagram, value)
+          return StatsD::Instrument::VOID
+        end
+
+        emit(precompiled_datagram.to_datagram(value))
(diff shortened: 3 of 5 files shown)


lib/statsd/instrument/compiled_metric.rb
```diff
@@ -177,6 +177,7 @@ def define_dynamic_method(tags)
           default_val = default_value
           default_val_assignment = default_val.nil? ? "" : " = #{default_val.inspect}"
           allow_block = allow_measuring_latency
+          client_emit_method = allow_block ? "emit_presampled_precompiled_#{method}_metric" : "emit_precompiled_#{method}_metric"
```
pedro-stanaka, line 183: Not blocking, but this code is subtle enough that documenting the invariant near the ternary would make future changes safer.

```suggestion
          # Block-enabled compiled metrics run through generate_block_handler first,
          # which applies sample_rate before emitting. Use the presampled client helper
          # here so distributions don't make a second sampling decision.
          client_emit_method = allow_block ? "emit_presampled_precompiled_#{method}_metric" : "emit_precompiled_#{method}_metric"
```

pedro-stanaka (approved):
LG - Apart from the comment, it would be nice if the sampling test was probabilistic since the beginning. It would have covered this. 

I think the mocked test is enough for now, you may want to bump the version already to avoid a PR just for the release.

PR #423, opened 2026-07-21
Add positional aggregator interface

pedro-stanaka:
## Why

Aggregation is currently hard-coded to the built-in implementation and receives metric options as keyword arguments. The aggregator is an internal hot-path interface, so a single fixed-arity positional contract is simpler, cheaper to dispatch, and easier to implement with alternative aggregation strategies.

## What changed

- Make the internal aggregator interface positional-only:

  ```ruby
  increment(name, value, tags, no_prefix, sample_rate)
  gauge(name, value, tags, no_prefix)
  aggregate_timing(name, value, tags, no_prefix, type, sample_rate)
  ```

- Keep the public `Client` metric API unchanged, including its keyword options.
- Add optional `aggregator:` injection to `Client.new` and `.from_env`.
- Preserve an injected aggregator through `with_options` / `clone_with_options` unless explicitly overridden.
- Providing `aggregator:` implicitly enables aggregation.
- Convert untagged aggregation calls to one shared frozen empty array; other tag values are forwarded as-is, so custom implementations may require canonical arrays.
- Keep sampling in `Client` before aggregation dispatch.
- Keep the existing precompiled aggregation methods and `flush` as part of the aggregator contract.
- Document the interface and add a changelog entry.

## Compatibility

Public `StatsD` and `Client` metric methods keep their existing signatures and behavior. Code that directly calls or prepends the internal `Aggregator` methods must update to the positional signatures.

Injected aggregators own prefix/default-tag policy and lifecycle behavior. Cloned clients reuse the injected instance unless `aggregator:` is overridden.

## Performance

Linux arm64, Ruby 3.4.10 + YJIT, 1M calls:

| Path | Median | Ruby allocations/call |
|---|---:|---:|
| Direct positional aggregator method | 24.6 ns/call | 0 |
| Client → aggregator | 32.5 ns/call | 0 |

A five-process before/after run found no regression in existing direct or CompiledMetric paths. Representative medians:

| Path | main | this branch |
|---|---:|---:|
| Client distribution | 1,351.9 ns | 1,321.8 ns |
| Dynamic compiled distribution | 184.5 ns | 178.7 ns |
| Static compiled distribution | 133.7 ns | 134.5 ns |
| Dynamic compiled counter | 216.6 ns | 214.2 ns |

## Validation

- `bundle exec rake test` — 397 runs, 1,111 assertions, 0 failures/errors
- `bundle exec rubocop` — 79 files, no offenses
- Ruby 2.7–3.3, ruby-head, JRuby, and TruffleRuby CI coverage
- UDP integration test

#gsd:52109

### DIFF
--- a/lib/statsd/instrument/aggregator.rb
+++ b/lib/statsd/instrument/aggregator.rb
@@ -148,7 +148,7 @@ module StatsD
       # @param tags [Hash{String, Symbol => String},Array<String>] The tags to attach to the counter.
       # @param no_prefix [Boolean] If true, the metric will not be prefixed.
       # @return [void]
-      def increment(name, value = 1, tags: [], no_prefix: false, sample_rate: CONST_SAMPLE_RATE)
+      def increment(name, value, tags, no_prefix, sample_rate)
         unless thread_healthcheck
           @sink << datagram_builder(no_prefix: no_prefix).c(name, value, sample_rate, tags)
           return
@@ -161,12 +161,6 @@ module StatsD
           @aggregation_state[key] ||= 0
           @aggregation_state[key] += value
         end
-      end
-
-      # Positional-only counter entrypoint used by external aggregation backends.
-      # @api private
-      def record_counter(name, value, tags, no_prefix, sample_rate)
-        increment(name, value, tags: tags, no_prefix: no_prefix, sample_rate: sample_rate)
       end
 
       # Aggregates a precompiled metric that can be combined into a single scalar for later flushing.
@@ -230,7 +224,7 @@ module StatsD
         end
       end
 
-      def aggregate_timing(name, value, tags: [], no_prefix: false, type: DISTRIBUTION, sample_rate: CONST_SAMPLE_RATE)
+      def aggregate_timing(name, value, tags, no_prefix, type, sample_rate)
         unless thread_healthcheck
           @sink << datagram_builder(no_prefix: no_prefix).timing_value_packed(
             name, type.to_s, [value], sample_rate, tags
@@ -255,13 +249,7 @@ module StatsD
         do_flush(aggregation_state) if aggregation_state
       end
 
-      # Positional-only timing entrypoint used by external aggregation backends.
-      # @api private
-      def record_timing(name, value, tags, no_prefix, type, sample_rate)
-        aggregate_timing(name, value, tags: tags, no_prefix: no_prefix, type: type, sample_rate: sample_rate)
-      end
-
-      def gauge(name, value, tags: [], no_prefix: false)
+      def gauge(name, value, tags, no_prefix)
         unless thread_healthcheck
           @sink << datagram_builder(no_prefix: no_prefix).g(name, value, CONST_SAMPLE_RATE, tags)
           return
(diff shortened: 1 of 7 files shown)


lib/statsd/instrument/aggregator.rb
```diff
@@ -148,7 +148,7 @@ def initialize(
       # @param tags [Hash{String, Symbol => String},Array<String>] The tags to attach to the counter.
       # @param no_prefix [Boolean] If true, the metric will not be prefixed.
       # @return [void]
-      def increment(name, value = 1, tags: [], no_prefix: false, sample_rate: CONST_SAMPLE_RATE)
+      def increment(name, value, tags, no_prefix, sample_rate)
```
kelhusseiny, line 151: shopify/core prepends all three changed methods with the old keyword signatures, at 

```ruby
//areas/core/shopify/components/platform/essentials/lib/patches/statsd_metric_default_tag.rb:80-96
``` 
(verified present on main):
```
def increment(name, value = 1, tags: [], no_prefix: false, sample_rate: 1.0)
  tags = StatsdMetricDefaultTag.add_default_tags(name, tags)
  super(name, value, tags: tags, no_prefix: no_prefix, sample_rate: sample_rate)
end
StatsD::Instrument::Aggregator.prepend(StatsdMetricDefaultTag::Patch)
```
I ran the PR branch with that patch shape prepended:
```
increment: ArgumentError: wrong number of arguments (given 5, expected 1..2)
gauge:     ArgumentError: wrong number of arguments (given 4, expected 2; required keyword: no_prefix)
parity test passes? false
  original: [[:req,:name],[:req,:value],[:req,:tags],[:req,:no_prefix],[:req,:sample_rate]]
  patch:    [[:req,:name],[:opt,:value],[:key,:tags],[:key,:no_prefix],[:key,:sample_rate]]
```
so since this a breaking change can we at least bump this as a major version upgrade?

test/client_test.rb
```diff
@@ -102,6 +102,55 @@ def test_client_from_env_with_aggregation
...
+
+    datagram = sink.datagrams.find { |d| d.name == "compiled.requests" }
+    assert_equal(2, datagram.value)
+  ensure
+    StatsD.singleton_client = old_client
+    client&.instance_variable_get(:@aggregator)&.instance_variable_get(:@flush_thread)&.kill
+  end
+
+  def test_clone_with_options_preserves_aggregator
+    sink = StatsD::Instrument::CaptureSink.new(parent: StatsD::Instrument::NullSink.new)
```
kelhusseiny, line 155: can we have a test where we get the aggregator injected? I don't see a test coverage for that

kelhusseiny (changes requested):
the breaking change is the only blocker I guess, if you're bumping this to a new major version that it'd be ok

Also can we make a draft PR on SFR using the new version and make it run through the verifier? I'd like to make sure it's safe to bump from functional and performance perspective (Verifier will automatically run a performance test)

lib/statsd/instrument/aggregator.rb
pedro-stanaka: Yeah, I thought about that. The aggregator api is internal though, that is why I was bumping minor. We are several versions behind in that app also. But I will bump the major to avoid any tripwires.

pedro-stanaka: That app is kind of a snow flake, and I am planning on replacing that monkey path with #427

test/client_test.rb
pedro-stanaka: Added.

pedro-stanaka:
Please check again @kelhusseiny

kelhusseiny (approved):

PR #429, opened 2026-09-18
Normalize newline characters in metric names and tags

pedro-stanaka:
## Summary

Extend the existing normalization to handle `\n` and `\r`, keeping each metric on one line.

- Replace newlines in names, prefixes, and set values with `_`.
- Keep the existing tag conventions: remove them from array/hash tags and compiled tags; use `_` in already-serialized string tags.
- Cover ordinary, aggregated, and compiled metrics, including default and dynamic tags.
- Skip replacement when no matching character is present. No new logging, counters, or configuration.
- Prepare patch release **4.0.1**: bump `VERSION` and move the changelog entry into its release section.

## Validation

- `bundle exec rake lint test`: full lint and all 395 tests pass on Ruby 3.4.5, both with the local dependency set and CI's RuboCop 1.91.0 / rubocop-shopify 3.1.0.
- Removed redundant escapes in the existing datagram parsers and replaced a version-dependent test lint suppression with explicitly escaped interpolation; parser behavior is unchanged.
- `gem build statsd-instrument.gemspec` successfully builds version 4.0.1 locally; nothing published.
- CI lint and all eight Ruby test jobs pass on `ad736bd`.

## Performance

Measured with a local, uncommitted harness on Apple M3 Pro, Ruby 3.4.5 + YJIT, NullSink (serialization/aggregation only, no network I/O). Medians of five baseline/candidate pairs, alternating order, 1,000,000 calls per case after warmup; baseline `d5b105f`.

| Valid-input case | Main ns/call | This branch ns/call | Allocations/call, both |
| --- | ---: | ---: | ---: |
| Client, no per-call tags | 314 | 311 | 2 |
| Client, four array tags | 712 | 688 | 3 |
| Client, 16 long array tags | 3,561 | 3,476 | 3 |
| Client, four hash tags | 979 | 930 | 6 |
| Set value | 258 | 280 | 1 |
| Compiled, static | 257 | 257 | 3 |
| Compiled, cached dynamic tags | 669 | 654 | 5 |
| Aggregated counter | 1,510 | 1,461 | 5 |
| Aggregated compiled counter | 286 | 294 | 0 |

No additional allocations for clean input. Most timing differences are within run-to-run variation; the set path adds a previously absent value scan, measured at about +22 ns/call (+8.5%). Inputs containing newlines now pay for normalization: a name measured 343 ns/call versus 203 ns previously, and a tag 575 ns versus 395 ns previously (the baseline left those newlines unchanged).


lib/statsd/instrument/client.rb
```diff
@@ -213,7 +213,7 @@ def initialize(
       #
       #   - We recommend using `snake_case.metric_names` as naming scheme.
       #   - A `.` should be used for namespacing, e.g. `foo.bar.baz`
-      #   - A metric name should not include the following characters: `|`, `@`, and `:`.
+      #   - A metric name should not include `|`, `@`, `:`, or newlines (`\n`, `\r`).
```
zcei, line 216: I wonder whether we should substitute any whitespace character (`\s+`) in metric names. Quick search revealed that at least the original implementation did this and e.g. a metric name of `prefix. my value` looks odd at least, and potentially breaks receivers.

lib/statsd/instrument/compiled_metric.rb
```diff
@@ -302,7 +302,7 @@ def build(name:, type:, client_prefix:, no_prefix:, default_tags:, static_tags:,
           # @param name [String] The metric name
           # @return [String] The normalized metric name
           def normalize_name(name)
-            name.tr(":|@", "_")
+            /[:|@\r\n]/.match?(name) ? name.tr(":|@\r\n", "_") : name.dup
```
zcei, line 305: Can't we just return the same `name` instance, or would this lead to problems down the way?
If the contract is that it always returns a new string, we might want to capture that in it's RBDoc comment block (or is it called JSDoc for Ruby as well? 😆)

lib/statsd/instrument/compiled_metric.rb
```diff
@@ -314,15 +314,16 @@ def normalize_name(name)
           def build_prefix(client_prefix, no_prefix)
             return "" if no_prefix || client_prefix.nil?
 
-            "#{client_prefix}."
+            prefix = "#{client_prefix}."
+            /[\r\n]/.match?(prefix) ? prefix.tr("\r\n", "_") : prefix
```
zcei, line 318: Should we just get ourselves a small sanitization helper that does it everywhere for us? e.g. I'm wondering here why we only care for the whitespace but not the other disallowed chars

zcei (commented):
The overall intention makes sense to me, though some clarifying questions before approving

lib/statsd/instrument/client.rb
pedro-stanaka: So, i checked other libraries they dont usually do this for all blank spaces, but I think it makes sense, so I implemented it.

lib/statsd/instrument/compiled_metric.rb
pedro-stanaka: Good catch! The `.dup` was there because define calls normalize_name(name).freeze. Returning the original string there would freeze the caller’s input too.

But that doesn’t need to be the sanitizer’s responsibility. I’ll have it return the same instance when nothing needs changing, and use -normalize_name(name) when storing the compiled name. That gives us a frozen string without freezing the caller’s mutable input.

 I’ll document that contract and keep tests covering both instance reuse and caller mutability. And it’s YARD
 here 😄

pedro-stanaka:
@zcei thanks for the review. Pls take a look again.

zcei (approved):
Awesome, the sanitization helper naming immediately made it clear why we only tested for a subset of the pattern 😆 🫶

